PR #11, opened 2016-04-06
Added tests for creating participants as well as updating/retrieving,…

alxdarksage:
… and tests for external IDs API.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsTest.java
@@ -48,9 +48,8 @@ public class ExternalIdsTest {
             identifiers.add(prefix+RandomStringUtils.randomAlphabetic(10));
         }
         DeveloperClient client = developer.getSession().getDeveloperClient();
+        client.addExternalIds(identifiers);
         try {
-            client.addExternalIds(identifiers);
-            
             PagedResourceList<ExternalIdentifier> page1 = client.getExternalIds(null, PAGE_SIZE, prefix, null);
             assertEquals(PAGE_SIZE, page1.getItems().size());
             assertEquals(LIST_SIZE, page1.getTotal());
@@ -77,10 +76,9 @@ public class ExternalIdsTest {
             assertEquals(0, page1.getTotal());
         } finally {
             client.deleteExternalIds(identifiers);
-
-            PagedResourceList<ExternalIdentifier> page = client.getExternalIds(null, null, prefix, null);
-            assertEquals(0, page.getTotal());
         }
+        PagedResourceList<ExternalIdentifier> page = client.getExternalIds(null, null, prefix, null);
+        assertEquals(0, page.getTotal());
     }
     
 }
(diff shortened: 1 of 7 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsTest.java
```diff
@@ -0,0 +1,86 @@
...
+        // with a random prefix that we'll use in all the queries.
+        String prefix = RandomStringUtils.randomAlphabetic(4);
+        
+        List<String> identifiers = Lists.newArrayListWithCapacity(LIST_SIZE);
+        for (int i=0; i < LIST_SIZE; i++) {
+            identifiers.add(prefix+RandomStringUtils.randomAlphabetic(10));
+        }
+        DeveloperClient client = developer.getSession().getDeveloperClient();
+        try {
+            client.addExternalIds(identifiers);
```
DwayneJengSage: Minor: We should move the addExternalIds() call before the try. Otherwise, if addExternalIds() fail, then deleteExternalIds will also fail, and that secondary failure will mask the original failure.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsTest.java
```diff
@@ -0,0 +1,86 @@
...
+            assertEquals(0, page1.getTotal());
+            
+            // different idFilter test (than the use of idFilter in all the tests. This time nothing should match.
+            page1 = client.getExternalIds(null, PAGE_SIZE, RandomStringUtils.randomAlphabetic(5), null);
+            assertEquals(0, page1.getTotal());
+        } finally {
+            client.deleteExternalIds(identifiers);
+
+            PagedResourceList<ExternalIdentifier> page = client.getExternalIds(null, null, prefix, null);
+            assertEquals(0, page.getTotal());
```
DwayneJengSage: Similarly, we should move the check that getExternalIds() returns nothing to after the finally block. This way, if the test fails, we're unlikely to have a secondary failure in the finally block that masks the original failure.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantsTest.java
```diff
@@ -89,38 +93,82 @@ public void cannotSetBadPageSize() {
...
-        assertEquals("First name", updated.getFirstName());
-        assertEquals("Last name", updated.getLastName());
-        assertEquals("external ID", updated.getExternalId());
-        assertEquals(SharingScope.ALL_QUALIFIED_RESEARCHERS, updated.getSharingScope());
-        assertEquals(true, updated.isNotifyByEmail());
-        assertEquals(Sets.newHashSet("group1"), updated.getDataGroups());
-        assertEquals(languages, updated.getLanguages());
-        assertEquals(attributes.get("phone"), updated.getAttributes().get("phone"));
+            ResearcherClient client = researcher.getSession().getResearcherClient();
+            client.createStudyParticipant(participant);
```
DwayneJengSage: Similarly, the createStudyParticipant() call should be before the try

DwayneJengSage:
Overall looks good. Some minor comments.

PR #19, opened 2016-04-29
Updating integration tests

alxdarksage:
Updating integration tests based on removal of user profile and its replacement with more general purpose API endpoints.


src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
```diff
@@ -0,0 +1,104 @@
...
+        developer = TestUserHelper.createAndSignInUser(UserParticipantTest.class, true, Roles.DEVELOPER);
+    }
+
+    @After
+    public void after() {
+        developer.signOutAndDeleteUser();
+    }
+
+    @Test
+    @Ignore
```
DwayneJengSage: Unignore these tests.

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
```diff
@@ -0,0 +1,104 @@
...
+        client.saveStudyParticipant(participant);
+        
+        participant = client.getStudyParticipant();
+        assertEquals(dataGroups, participant.getDataGroups());
+        
+        participant = new StudyParticipant.Builder().withDataGroups(Sets.<String>newHashSet()).build();
+        client.saveStudyParticipant(participant);
+        
+        participant = client.getStudyParticipant();
+        assertTrue(participant.getDataGroups().isEmpty());
```
DwayneJengSage: Verify data groups in session are updated?

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+
+/**
+ * Test of the participant APIs that act on the currently authenticated user, which have replaced the 
+ * user profile APIs as well as individual calls to update things like dataGroups, externalId, or 
+ * sharing settings.
+ *
+ */
+@Category(IntegrationSmokeTest.class)
```
DwayneJengSage: The IntegrationSmokeTest annotation makes this test run in prod.

Do we have a guideline for what tests should run in prod? I feel like tests that indicate general server healthiness and inexpensive tests should live in prod. Functional validation tests and heavy-weight tests should live in staging. I'm not sure what category this falls under.

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
```diff
@@ -0,0 +1,104 @@
...
+    private TestUser developer;
+
+    @Before
+    public void before() {
+        developer = TestUserHelper.createAndSignInUser(UserParticipantTest.class, true, Roles.DEVELOPER);
+    }
+
+    @After
+    public void after() {
+        developer.signOutAndDeleteUser();
```
DwayneJengSage: Minor: If creating the user fails, this will throw a null pointer exception. Should put a null-check around the signOutAndDeleteUser() call.

DwayneJengSage:
Overall looks good. A few minor comments. Also, you need to resolve merge conflicts.

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
alxdarksage: Removed @Ignore

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
alxdarksage: I think this test is running in production because the usr profile APIs are an API that all the tests are definitely using, and we wanted to verify. The tests in this class are using the "self" APIs of the participants APIs, and fall in this category. the ParticipantsTest class runs tests against the APIs for researchers and those don't run in production. 

I would lean on running these in production but I don't feel strongly about it and we could remove them.

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserParticipantTest.java
alxdarksage: Added to all @After/@AfterClass checks where it wasn't being done.

DwayneJengSage:
Integration tests look good. Still need to merge conflicts.

PR #22, opened 2016-05-05
add appVersion to integ tests

DwayneJengSage:
Add appVersion to Upload Test Integ Tests

Older versions of the Upload API ignore app version. New versions of the API parse it and expect it to be there. The new API won't error out if appVersion is missing, but it will log a message in the upload validation status, which the integ tests validate. This change is compatible with both the old API and the new one.

Tested with the new change (in  my local server) and against the old code in devo, staging, and prod.


DwayneJengSage:
Merging pull request, since it's all binary files anyway.

PR #28, opened 2016-05-13
Adding tests of session changes, including session updates on certain calls (saving participant).

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/TestUserHelper.java
```diff
@@ -23,70 +23,95 @@
...
-    public static TestUser createAndSignInUser(Class<?> cls, boolean consent, Roles... roles) {
+    public static TestUser createAndSignInUser(Class<?> cls, boolean consent) {
+        return createAndSignInUser(cls, consent, Roles.TEST_USERS);
+    }
+    
+    public static TestUser createAndSignInUser(Class<?> cls, boolean consent, Roles role) {
+        StudyParticipant participant = new StudyParticipant.Builder()
+                .withRoles(Sets.newHashSet(role)).build();
+        return createAndSignInUser(cls, consent, participant);
+    }
```
DwayneJengSage: Is it worth keeping the varargs Roles method? This trivially supports both the no roles and the one role helper method and allows us to trivially support a test with multiple roles, if we ever need it.

alxdarksage: Added that back and removed one of the two method.

PR #39, opened 2016-06-06
Switching to the user of the resource-specific clients.

alxdarksage:


DwayneJengSage:
I'm going ahead and merging this because it's blocking me right now.

PR #40, opened 2016-06-07
Streamline UploadTest, add new schema field types

DwayneJengSage:
Now that we have extensive unit tests in BridgePF, we can streamline some of the tests in UploadTest. Also add the new Upload v2 fields to ensure we're testing those.

See also https://github.com/Sage-Bionetworks/BridgePF/pull/1157 and https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/285

There's a lot of files being changed in this change list, but almost all of them are test data changes to match the new test schemas.


src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
```diff
@@ -166,60 +115,13 @@ public static void deleteUser() {
...
-        testUpload("non-json-encrypted");
-    }
-
-    @Test
-    public void appVersionAboveMax() throws Exception {
-        testUpload("max-app-version-test-v30-encrypted");
-    }
-
-    @Test
-    public void appVersionBelowMax() throws Exception {
```
alxdarksage: Do we still need nonJson, appVersionBelow/AboveMax tests, did they go somewhere else?

PR #46, opened 2016-06-16
Add tests to ensure these email-based methods don't blow up.

alxdarksage:
Also verified in logging and through the researcher UI, and in the bridge testing account where these email were sent to.


DwayneJengSage:
Looks good.

PR #50, opened 2016-06-23
Adding tests of withdrawing from all consents.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
```diff
@@ -212,4 +213,30 @@ public void canEmailConsentAgreement() {
...
+        TestUser user = TestUserHelper.createAndSignInUser(ConsentTest.class, true);
+        try {
+            // Can get activities without an error... user is indeed consented.
+            user.getSession().getUserClient().getScheduledActivities(1, DateTimeZone.UTC);
+            for (ConsentStatus status : user.getSession().getConsentStatuses().values()) {
+                assertTrue(status.isConsented());
+            }
+            
+            user.getSession().getUserClient().withdrawAllConsentsToResearch("I'm just a test user.");
+            user.signOut();
```
DwayneJengSage: The Java SDK sets the sharing scope and consents in the user's session when they call withdrawAll(). Should we test for that before we sign the user out?

DwayneJengSage:
Overall looks good. One comment about sessions. Also it looks like there's a merge conflict that needs to be resolved.

alxdarksage:
Added session check to tests. Merging to fix dev integration.

PR #76, opened 2016-08-29
Minor schema test changes in preparation of upcoming mutable schema changes

DwayneJengSage:
There are a few minor changes in mutable schema validation. Before we push those changes server-side, we need to make some adjustments to the integ tests. In particular:
- Split the schema test for "optional values" to create separate fields for various field properties, instead of one mega-field with all properties.
- Added fields must have minAppVersion specified.

These are technically breaking changes, but given there are no automated processes that use the new field properties or update schema revisions in place, these changes should be safe to make.

Testing done:
- Ran against local server to verify compatibility with changes.
- Ran against dev to verify compatibility with old code.

### DIFF
--- a/src/test/resources/logback-test.xml
+++ b/src/test/resources/logback-test.xml
@@ -7,9 +7,9 @@
     </encoder>
   </appender>
 
-  <logger name="org.sagebionetworks.bridge.sdk.BaseApiCaller" level="debug" />
+  <logger name="org.sagebionetworks.bridge.sdk.BaseApiCaller" level="warn" />
 
-  <root level="debug">
+  <root level="info">
     <appender-ref ref="STDOUT" />
   </root>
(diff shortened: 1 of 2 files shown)


src/test/resources/logback-test.xml
```diff
@@ -7,9 +7,9 @@
     </encoder>
   </appender>
 
-  <logger name="org.sagebionetworks.bridge.sdk.BaseApiCaller" level="debug" />
+  <logger name="org.sagebionetworks.bridge.sdk.BaseApiCaller" level="warn" />
```
alxdarksage: I like setting logging to debugging because when there’s an error, you can see the JSON traffic sent back and forth to the server, and it’s tedious to change and redeploy it. But also fine to change this back to warning

PR #82, opened 2016-09-08
fix mutable schema tests; add survey schema tests

DwayneJengSage:
Mutable schema tests didn't do anything meaningful. I fixed them to actually modify the schema.

Added tests for survey schema tests. These didn't exist before. Now they do.

See also https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/309

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveySchemaTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveySchemaTest.java
@@ -1,0 +1,320 @@
+package org.sagebionetworks.bridge.sdk.integration;
+
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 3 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveySchemaTest.java
```diff
@@ -0,0 +1,320 @@
...
+    public void deleteSchemas() {
+        try {
+            schemaClient.deleteUploadSchemaAllRevisions(surveyId);
+        } catch (EntityNotFoundException ex) {
+            // Suppress the exception, as the test may have already deleted the schema.
+        }
+    }
+
+    @AfterClass
+    public static void deleteResearcher() {
```
alxdarksage: deleteDeveloper()

PR #86, opened 2016-10-11
add new integ test for calling update record exporter status

zaclinsage:
add new integ test for calling update record exporter status


pom.xml.versionsBackup
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
alxdarksage: I think remove this file maybe add to .gitignore as well.

alxdarksage (approved):
I would remove one file, otherwise looks good.

PR #92, opened 2016-10-25
Updating a few more test cases.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
```diff
@@ -182,4 +241,19 @@ public static Study getStudy(String identifier, Long version) {
         return set;
     }
     
+    public static <T> boolean assertListsEqualIgnoringOrder(List<T> list1, List<T> list2) {
```
liujoshua: If the lists have duplicates the behavior might not be as expected. Should [1,1,2] equal [1,2,2]? In this case a javadoc would helpful

liujoshua (approved):

src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
alxdarksage: I'll convert to sets first...

liujoshua (approved):

PR #108, opened 2016-12-09
pass all integ test

zaclinsage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyTest.java
```diff
@@ -22,38 +24,166 @@
...
+    private void loadProperties(final String fileName, final Properties properties) {
+        File file = new File(fileName);
+        if (file.exists()) {
+            try (InputStream in = new FileInputStream(file)) {
+                properties.load(in);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
```
alxdarksage: How will this load the values from environment variables that we pass in on the command line when using Jenkins?

Are there values for these when testing locally? Can you send them to us via email (not git) so we can run the integration tests locally?

alxdarksage (changes requested):

PR #113, opened 2016-12-14
Test to verify auto login

liujoshua:


src/test/java/org/sagebionetworks/bridge/sdk/integration/SessionRefreshTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test;
+
+import org.sagebionetworks.bridge.rest.api.ParticipantsApi;
+
+public class SessionRefreshTest {
+    private static TestUserHelper.TestUser user;
+
+    @BeforeClass
+    public static void createUser() throws Exception {
+        user = TestUserHelper.createAndSignInUser(SessionRefreshTest.class, false);
```
alxdarksage: To avoid 412 here, just make a developer.

PR #121, opened 2017-01-26
modify tests relate to get uploads and add pagination test as well

zaclinsage:


pom.xml
```diff
@@ -49,14 +49,20 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.12.23</version>
```
alxdarksage: bump to 0.12.26

pom.xml
```diff
@@ -49,14 +49,20 @@
...
             <groupId>org.sagebionetworks</groupId>
             <artifactId>synapseJavaClient</artifactId>
             <version>161.0-4-g843de2c</version>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>1.7.21</version>
+            <scope>test</scope>
+        </dependency>
```
alxdarksage: This is included in the rest client, is that not included as part of the tests? I think Joshua just removed logging dependencies from these tests

src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyTest.java
```diff
@@ -375,13 +388,38 @@ public void getStudyUploads() throws Exception {
...
             assertEquals(endTime, results.getEndTime());
 
             assertEquals(count+2, results.getItems().size());
             assertNotNull(getUpload(results, uploadSession.getId()));
             assertNotNull(getUpload(results, uploadSession2.getId()));
+
+            // then test pagination by setting max pagesize to 1
+            UploadList pagedResults = studiesApi.getUploads(startTime, endTime, 1L, null).execute().body();
+            assertEquals(startTime, pagedResults.getStartTime());
+            assertEquals(endTime, pagedResults.getEndTime());
```
alxdarksage: Add test for page size to be null... this is the test that will verify the BridgePF routes file allows the parameter to be optional...

alxdarksage (changes requested):
A couple of minor changes. Also, do we have a test to verify that an admin can logically delete a study, then it doesn't appear in the APIs, then the admin developer can re-enable it? As much of that logic that is in the API, we should have tests for.

PR #125, opened 2017-02-03
fix StudyTest, SignUpTest

DwayneJengSage:
Changes:
* One test in SignUpTest wasn't cleaning up accounts.
* One test in StudyTest was using an admin account to call a Developer-only API.
* Nonsensical counts in one test in StudyTest.
* References to getTotal() for getStudyUploads() for StudyTest removed, as this value isn't particularly meaningful.
* Pagination tests for getStudyUploads() assume that the uploads created by the test are the only uploads. Removed this assumption. We already verify the uploads we're looking for in another part of the test.

Tested in both Prod (old code) and Staging (new code).


zaclinsage (commented):
GTG

PR #126, opened 2017-02-04
add integ test to catch missing ReportIndex issue

DwayneJengSage:
Integ test to catch and verify missing ReportIndex issue. See https://github.com/Sage-Bionetworks/BridgePF/pull/1365 for more details.

Tested this against Staging and verified that it fails. Tested this against my local changes and verified that it succeeds.


alxdarksage:
This was supposed to be merged, not closed, my mistake.

PR #135, opened 2017-03-08
integ tests for Compound Activities

DwayneJengSage:
Pre-reqs:
* These integ tests caught bugs fixed by https://github.com/Sage-Bionetworks/BridgePF/pull/1386 Those fixes should be checked in and pushed out before this is checked in.
* This depends on JavaSDK changes https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/379

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/CompoundActivityDefinitionTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/CompoundActivityDefinitionTest.java
@@ -123,7 +123,7 @@ public class CompoundActivityDefinitionTest {
             // Test list. Since there might be other defs from other tests, page through the defs to find the ones
             // corresponding to the test.
             List<CompoundActivityDefinition> defList = compoundActivityDefinitionsApi
-                    .getAllCompoundActivityDefinitionsInStudy().execute().body().getItems();
+                    .getAllCompoundActivityDefinitions().execute().body().getItems();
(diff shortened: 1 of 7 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/CompoundActivityDefinitionTest.java
```diff
@@ -0,0 +1,168 @@
...
+                compoundActivityDefinitionsApi.deleteCompoundActivityDefinition(taskId2).execute();
+            } catch (RuntimeException ex) {
+                // squelch error
+            }
+        }
+    }
+
+    // We can't set the "type" field for schema objects we create programmatically. However, the rest-client can. This
+    // means we can't use .equals(). This is a helper method to make validation less insane.
+    private static void assertSchemaList(List<SchemaReference> schemaList) {
```
alxdarksage: I took a different tack and used reflection to set type so I could test equality:
Tests.setVariableValueInObject(plan, "type", "SchedulePlan");

alxdarksage (approved):

src/test/java/org/sagebionetworks/bridge/sdk/integration/CompoundActivityDefinitionTest.java
DwayneJengSage: This works for type, but not for href. The value of href is server-generated, so it's impractical to get equals() to work here.

PR #137, opened 2017-03-16
add codes for disable export

zaclinsage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyTest.java
```diff
@@ -244,6 +244,9 @@ public void crudStudy() throws Exception {
         // And this is also true
         assertTrue("emailVerificationEnabled should be true", newStudy.getEmailVerificationEnabled());
 
+        // assert disable study
+        assertFalse(newStudy.getDisableExport());
```
DwayneJengSage: I don't think this test is sufficient. If disableExport defaults to false, and we check that it's false, how can we make sure that we're properly saving it to the server, and not just getting the default value back?

DwayneJengSage (changes requested):

DwayneJengSage:
Also, need to fix merge errors.

DwayneJengSage (changes requested):
It looks like after the merge, you need to bump the JavaSDK version again.

DwayneJengSage (approved):
Looks good. Will merge after JavaSDK builds.

PR #148, opened 2017-04-13
add non auth user get and query shared modules integ test

zaclinsage:
integ tests done;

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleMetadataTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleMetadataTest.java
@@ -77,7 +77,7 @@ public class SharedModuleMetadataTest {
     public void testNonAuthUserGetAndQueryCalls() throws Exception {
         // first create a test metadata
         SharedModuleMetadata metadataToCreate = new SharedModuleMetadata().id(moduleId).version(1)
-                .name(MODULE_NAME).schemaId(SCHEMA_ID).schemaRevision(SCHEMA_REV).published(true);
+                .name(MODULE_NAME).schemaId(SCHEMA_ID).schemaRevision(SCHEMA_REV);
         SharedModuleMetadata metadata = sharedDeveloperModulesApi.createMetadata(metadataToCreate).execute()
                 .body();
         // execute query and get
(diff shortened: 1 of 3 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/TestUserHelper.java
```diff
@@ -140,6 +143,22 @@ public static TestUser getSignedInUser(SignIn signIn) {
...
+    // return an non-auth user
+    public static <T> T getNonAuthClient(Class<T> service) {
+        String email = CONFIG.fromProperty(Config.Props.SHARED_DEVELOPER_EMAIL);
+        String password = CONFIG.fromProperty(Config.Props.SHARED_DEVELOPER_PASSWORD);
+        String studyId = CONFIG.fromProperty(Config.Props.SHARED_DEVELOPER_STUDY);
+        SignIn signIn = new SignIn().email(email).password(password).study(studyId);
+
+        // only used to get host url from client manager
+        ClientManager manager = new ClientManager.Builder().withSignIn(signIn).withConfig(CONFIG)
+                .withClientInfo(CLIENT_INFO).withAcceptLanguage(LANGUAGES).build();
```
DwayneJengSage: I think it would be better to just expose a public static method in ClientManager which takes in an Environment and returns the URL, instead of creating a dummy signIn that we just ignore.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleMetadataTest.java
```diff
@@ -164,6 +189,26 @@ private SharedModuleMetadata testCreateGet(int expectedVersion, Integer inputVer
         return createdMetadata;
     }
 
+    private SharedModuleMetadata testCreateGetPublished(int expectedVersion, Integer inputVersion) throws Exception {
+        // create
+        SharedModuleMetadata metadataToCreate = new SharedModuleMetadata().id(moduleId).version(inputVersion)
+                .name(MODULE_NAME).schemaId(SCHEMA_ID).schemaRevision(SCHEMA_REV).published(true);
+        SharedModuleMetadata createdMetadata = sharedDeveloperModulesApi.createMetadata(metadataToCreate).execute()
+                .body();
```
DwayneJengSage: I don't think it's necessary to create a generic helper method for this. It's also not necessary to make it published. Nor is it necessary to validate create() or get() as these are tested elsewhere. You're better off just using the 2 lines to create the module metadata directly in the test itself.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #149, opened 2017-04-13
add integ test for upload schema published flag in both upload schema…

zaclinsage:
… and import shared modules

integ tests done;

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleTest.java
@@ -123,6 +123,7 @@ public class SharedModuleTest {
         // Get local schema and verify some fields.
         localSchema = apiDeveloper.getClient(UploadSchemasApi.class).getUploadSchema(importStatus.getSchemaId(),
                 importStatus.getSchemaRevision().longValue()).execute().body();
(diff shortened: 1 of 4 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleTest.java
```diff
@@ -123,6 +123,7 @@ public void byIdAndVersionSchemaSuccess() throws Exception {
         // Get local schema and verify some fields.
         localSchema = apiDeveloper.getClient(UploadSchemasApi.class).getUploadSchema(importStatus.getSchemaId(),
                 importStatus.getSchemaRevision().longValue()).execute().body();
+        sharedSchema.setPublished(true);
```
DwayneJengSage: Similarly, this is also the wrong way to test this.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleTest.java
```diff
@@ -161,6 +162,7 @@ public void latestPublishedVersionSuccess() throws Exception {
...
```
DwayneJengSage: Same here.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #150, opened 2017-04-14
Adding some tests for email-based sign in.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthenticationTest.java
```diff
@@ -11,31 +11,41 @@
...
 
     private static TestUser testUser;
     private static AuthenticationApi authApi;
+    private static StudiesApi studiesApi;
     
     @BeforeClass
     public static void beforeClass() throws IOException {
-        testUser = TestUserHelper.createAndSignInUser(AuthenticationTest.class, true);
+        // User must be an admin to toggle emailSignInEnabled
+        testUser = TestUserHelper.createAndSignInUser(AuthenticationTest.class, true, Role.DEVELOPER, Role.ADMIN);
```
DwayneJengSage: Instead of making the test user an admin, wouldn't it be better to just use the admin user? This separates the admin account privileges from the test account under test.

pom.xml
```diff
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.12.46</version>
+            <version>0.12.47</version>
```
DwayneJengSage: Need to resolve merge conflict. See other comment in other pull requests about 12.49 vs 12.50.

DwayneJengSage (commented):
Two minor comments, one about merge conflict, one about permissions. Otherwise, this looks good.

src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthenticationTest.java
```diff
@@ -11,31 +11,45 @@
...
     private static TestUser testUser;
+    private static TestUser adminUser;
     private static AuthenticationApi authApi;
+    private static StudiesApi studiesApi;
+    private static StudiesApi adminStudiesApi;
     
     @BeforeClass
     public static void beforeClass() throws IOException {
-        testUser = TestUserHelper.createAndSignInUser(AuthenticationTest.class, true);
+        testUser = TestUserHelper.createAndSignInUser(AuthenticationTest.class, true, Role.DEVELOPER);
```
DwayneJengSage: I think there was a misunderstanding here. Originally, the test user had no special permissions. You originally added Developer and Admin permissions to be able to get and update the study. It looks like you moved the Admin permissions to the admin account, but now we have this weird test where the account that gets the study and the account that updates the study are two different accounts, and the user under test has Developer permissions to get the study.

It would be much cleaner if the user under test had no permissions and the admin account just did everything.

PR #152, opened 2017-04-18
add module ID and version to schemas and surveys; add them when importing a module

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-1798

See also:
* https://github.com/Sage-Bionetworks/BridgePF/pull/1426
* https://github.com/Sage-Bionetworks/BridgeDocs/pull/141
* https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/397
* https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/152

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.12.48</version>
+            <version>0.12.49</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


zaclinsage (commented):
Once we added this new feature, will it be easier to use getMetadataByIdAndVersion() instead of query for verifying existence of schema/survey?

DwayneJengSage:
Not necessarily. This annotates surveys and schemas in the local studies, not in the shared studies.

alxdarksage:
Also looks good. btw if you try and import into the shared study, what happens? Do we prevent it?

PR #182, opened 2017-08-17
minor fixes to ActivityEventTest

DwayneJengSage:
Three fixes:
* Test uses two consecutive calls to DateTime.now(), which can in very rare circumstances be off by a millisecond or two.
* Test does not clean up test accounts it creates.
* Test tags test accounts with wrong class names.


liujoshua (approved):
Thanks for fixing

PR #183, opened 2017-08-28
Integration tests for new scheduled activities history API.

alxdarksage:

PR #185, opened 2017-08-29
fix ScheduledActivityTest - unique activity labels instead of non-unique task IDs

DwayneJengSage:
ScheduledActivityTest uses task IDs to find its activities. However, task IDs aren't unique, which means this test sometimes conflicts with other tests.

Fixed the test to use unique activity labels, to remove the conflicts.


DwayneJengSage:
Note that this will conflict with https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/183

alxdarksage (approved):

PR #189, opened 2017-09-08
add surveyGuid/CreatedOn to Synchronous Health Data API

DwayneJengSage:
See also
* https://github.com/Sage-Bionetworks/BridgePF/pull/1550
* https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/432

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/HealthDataTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/HealthDataTest.java
@@ -19,12 +19,19 @@ import org.junit.experimental.categories.Category;
 import org.sagebionetworks.bridge.rest.RestUtils;
 import org.sagebionetworks.bridge.rest.api.HealthDataApi;
 import org.sagebionetworks.bridge.rest.api.ParticipantsApi;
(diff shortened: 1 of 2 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/HealthDataTest.java
```diff
@@ -133,4 +179,23 @@ public void test() throws Exception {
...
+
+        // submit and validate - Most of the record attributes are already validated in the previous test. Just
+        // validate survey ID was set properly as the schema ID and that the data is correct.
+        HealthDataRecord record = user.getClient(HealthDataApi.class).submitHealthData(submission).execute().body();
+        assertEquals(SURVEY_ID, record.getSchemaId());
+        assertNotNull(record.getSchemaRevision());
+
+        Map<String, String> returnedDataMap = RestUtils.toType(record.getData(), Map.class);
+        assertEquals(1, returnedDataMap.size());
+        assertEquals("C", returnedDataMap.get("answer-me"));
```
alxdarksage: The value as returned from the call Is correct, is there an API to retrieve data after uploaded? If so we should verify the data is also correct.

alxdarksage (approved):

PR #193, opened 2017-09-14
Adding a test for a sequence schedule (time-limited recurring schedule).

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduleTest.java
```diff
@@ -64,6 +66,40 @@ public void after() throws Exception {
...
+        planGuid = schedulesApi.createSchedulePlan(plan).execute().body().getGuid();
+        
+        ForConsentedUsersApi usersApi = user.getClient(ForConsentedUsersApi.class);
+        
+        String scheduledOn1 = LocalDate.now().toString() + "T14:00:00.000Z";
+        String scheduledOn2 = LocalDate.now().plusDays(1).toString() + "T14:00:00.000Z";
+        String scheduledOn3 = LocalDate.now().plusDays(2).toString() + "T14:00:00.000Z";
+        
+        ScheduledActivityList list = usersApi.getScheduledActivities("+00:00", 4, null).execute().body();
+        assertEquals(3, list.getItems().size());
```
DwayneJengSage: This test will fail if there are other schedules in the API study. (Which is very likely to happen if a previous test fails to clean up, or if there's another schedule from a manual test.) You should instead generate a random label for your activity, then filter the ScheduledActivityList for activities that match your label. See other tests in ScheduleTest or ScheduledActivityTest for examples.

DwayneJengSage (changes requested):

alxdarksage:
Updated based on review.

DwayneJengSage (approved):

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.sagebionetworks.bridge.rest.model.SchemaReference;
+import org.sagebionetworks.bridge.rest.model.Study;
+import org.sagebionetworks.bridge.rest.model.StudyParticipant;
+import org.sagebionetworks.bridge.rest.model.SurveyReference;
+import org.sagebionetworks.bridge.sdk.integration.TestUserHelper.TestUser;
+
+import com.amazonaws.services.identitymanagement.model.User;
+import com.google.common.collect.Lists;
+
+public class AppConfigTest {
```
DwayneJengSage: Should this have been split off into a different code review? It looks like we're bundling two different changelists when we don't need to be.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduleTest.java
```diff
@@ -64,6 +66,46 @@ public void after() throws Exception {
...
+        String randomLabel = Tests.randomIdentifier(ScheduleTest.class);
+        List<Activity> taggedActivities = Tests.labelActivities(acts, randomLabel);
+        
+        // I've made this shorter so we can verify it without paging, etc.
+        Schedule schedule = new Schedule();
+        schedule.setLabel("This is a squence schedule");
+        schedule.setScheduleType(ScheduleType.RECURRING);
+        schedule.setInterval("P1D");
+        schedule.setExpires("P1D");
+        //schedule.setSequencePeriod("P3D");
```
DwayneJengSage: It looks like this test still works even without a sequence period, which suggests to me that it's not testing sequence period at all. Should we tweak the test to make sure that it's testing what we think it's testing?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
```diff
@@ -0,0 +1,148 @@
...
+        study = developer.getClient(StudiesApi.class).getUsersStudy().execute().body();
+    }
+    
+    @After
+    public void after() throws Exception {
+        AppConfigList list = devApi.getAppConfigs().execute().body();
+        for (AppConfig config : list.getItems()) {
+            adminApi.deleteAppConfig(config.getGuid()).execute();
+        }
+    }
```
DwayneJengSage: Do we need to delete the users we've created?

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
alxdarksage: Yes, sorry. This should be here.

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
```diff
@@ -0,0 +1,148 @@
...
+        appConfig.setSchemaReferences(schemaReferences);
+        appConfig.setSurveyReferences(surveyReferences);
+        
+        // Create
+        GuidVersionHolder holder = devApi.createAppConfig(appConfig).execute().body();
+        appConfig.setGuid(holder.getGuid());
+        appConfig.setVersion(holder.getVersion());
+        
+        assertEquals(schemaRef1, appConfig.getSchemaReferences().get(0));
+        assertEquals(surveyRef1, appConfig.getSurveyReferences().get(0));
```
DwayneJengSage: You're comparing the object you created with itself, not with the result from the server.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #194, opened 2017-09-20
Upload Metadata

DwayneJengSage:
See https://github.com/Sage-Bionetworks/BridgePF/pull/1565 and https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/435


src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyTest.java
```diff
@@ -371,6 +376,119 @@ public void developerCannotChangeAdminOnlySettings() throws Exception {
...
+    private static void removeFieldDefByName(String fieldName, Study study) {
+        Iterator<UploadFieldDefinition> fieldDefIter = study.getUploadMetadataFieldDefinitions().iterator();
+        while (fieldDefIter.hasNext()) {
+            UploadFieldDefinition oneFieldDef = fieldDefIter.next();
+            if (oneFieldDef.getName().equals(fieldName)) {
+                fieldDefIter.remove();
+                break;
+            }
+        }
+    }
```
alxdarksage: Would it be easier to just check and add the fields that are needed for a test, then leave? Future tests could retrieve study, see it was configured correctly, and move on. We have already configured dataGroups this way in the API study.

alxdarksage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyTest.java
DwayneJengSage: No, because the whole point of the test is to test that we *can* add and remove fields.

alxdarksage (approved):

PR #206, opened 2017-10-25
Fix time zone so the two days don't have different time zones when crossing DST boundary.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -364,7 +364,8 @@ public void createSchedulePlanGetScheduledActivitiesV4() throws Exception {
         
         // If we ask for a date range that doesn't include it, it is not returned.
         list = participantsApi.getParticipantActivityHistory(user.getSession().getId(), activity.getGuid(),
-                DateTime.now().plusDays(10), DateTime.now().plusDays(12), null, 5).execute().body();
+                DateTime.now(DateTimeZone.UTC).plusDays(10), DateTime.now(DateTimeZone.UTC).plusDays(12), null, 5)
+                .execute().body();
```
DwayneJengSage: I’d recommend using a timezone that’s neither UTC nor PST/PDT, specifically so we can make sure we aren’t unintentionally encoding a timezone assumption somewhere in our code.

DwayneJengSage (approved):

PR #209, opened 2017-11-10
Phone signup signin

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthenticationTest.java
```diff
@@ -192,4 +198,34 @@ public void secondTimeSignUpLooksTheSameAsFirstTimeSignUp() throws Exception {
...
     }
+    
+    @Test(expected = InvalidEntityException.class)
+    public void requestPhoneSignInWithoutPhone() throws Exception {
+        AuthenticationApi authApi = testUser.getClient(AuthenticationApi.class);
+
+        PhoneSignInRequest phoneSignIn = new PhoneSignInRequest().study(testUser.getStudyId());
+
+        Response<Message> response = authApi.requestPhoneSignIn(phoneSignIn).execute();
+        assertEquals(202, response.code());
```
liujoshua: This line will never be reached. Just thought we'd remove it to keep it consistent with phoneSignInThrows()

PR #221, opened 2017-12-20
Fix OAuthTest to not interfere with FitBit Worker

DwayneJengSage:


alxdarksage (approved):

alxdarksage:
Can we push this out without bumping the version?

PR #230, opened 2018-02-01
Test for update identifiers API.

alxdarksage:


DwayneJengSage (changes requested):
Is it worth having the inverse test (add phone to user with email) to make sure the JavaSDK is set up correctly and we can create and end both types of requests?

alxdarksage:
Added a test to add a phone to an account with an email address

DwayneJengSage (approved):

PR #239, opened 2018-02-27
Update API test for retrieving app config.

alxdarksage:


DwayneJengSage (commented):
How easy is it to test the AppConfig API publicly (without a user)? If it's easy to do, we should change the test to do that.

If not, this is probably fine.

alxdarksage:
Good point. I changed the end of the test to use the public API and verified that it is filtering.

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
```diff
@@ -33,27 +36,21 @@
 
 public class AppConfigTest {
     
-    private TestUser user;
+    //private TestUser user;
```
DwayneJengSage: Minor: Remove commented out user. (And equivalent code in before() and after().

DwayneJengSage (approved):
One minor comment. Otherwise, looks good.

DwayneJengSage (approved):

PR #243, opened 2018-03-08
Removing code that's no longer necessary to send a useful UA header.

alxdarksage:


DwayneJengSage (changes requested):
I think the integ tests should clearly identify itself as integ tests. With that in mind, I think the app name should still say Integration Tests.

DwayneJengSage (approved):

PR #251, opened 2018-03-23
Phone verification

alxdarksage:
Going to merge this so the tests pass, then work on being able to update external ID in separate PR (it's small but worth treating separately).

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthenticationTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthenticationTest.java
@@ -63,6 +63,13 @@ public class AuthenticationTest {
         
         adminUser = TestUserHelper.getSignedInAdmin();
         adminStudiesApi = adminUser.getClient(StudiesApi.class);
+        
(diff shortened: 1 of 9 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthenticationTest.java
```diff
@@ -189,7 +221,7 @@ public void emailVerificationThrowsTheCorrectError() throws Exception {
         assertEquals(400, response.getStatusLine().getStatusCode());
         
         JsonNode node = new ObjectMapper().readTree(EntityUtils.toString(response.getEntity()));
-        assertEquals("Email verification token has expired (or already been used).", node.get("message").asText());
+        assertEquals("Verification token has expired (or already been used).", node.get("message").asText());
```
DwayneJengSage: Should we have an equivalent test for verifyPhone?

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #259, opened 2018-04-16
 Test of signing up and signing in with external ID.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
```diff
@@ -0,0 +1,122 @@
...
+        changeExternalIdValidation(false);
+        developer = TestUserHelper.createAndSignInUser(ExternalIdSignUpTest.class, false, Role.DEVELOPER);
+        extIdsClient = developer.getClient(ExternalIdentifiersApi.class);
+        authClient = developer.getClient(AuthenticationApi.class);
+        
+        ExternalIdentifierList list = extIdsClient.getExternalIds(null, 50, null, null).execute().body();
+        if (!list.getItems().isEmpty()) {
+            extIdsClient.deleteExternalIds(list.getItems().stream()
+                    .map(ExternalIdentifier::getIdentifier).collect(Collectors.toList())).execute();
+        }
```
DwayneJengSage: This conflicts with local testing. It would be better to create a randomized external ID instead.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
```diff
@@ -0,0 +1,122 @@
...
+        for (ExternalIdentifier identifier : list.getItems()) {
+            if ("AAA".equals(identifier.getIdentifier())) {
+                assertTrue(identifier.getAssigned());        
+            }
+        }
+        
+        // Prove you can sign in with this account (status = enabled)
+        SignIn signIn = new SignIn().externalId("AAA").study(Tests.STUDY_ID).password(Tests.PASSWORD);
+        try {
+            authClient.signInV4(signIn).execute().body();
```
DwayneJengSage: Need to add a fail().

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
alxdarksage: Changed this to create and delete to random external IDs.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
alxdarksage: Added.

alxdarksage:
Also fixed a permission issue with the clients that became apparent once the tests were changed.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            // Enable validation, and an account with just an external ID will succeed
+            changeExternalIdValidation(true);
+            devIdsClient.addExternalIds(ImmutableList.of(externalId, otherExternalId)).execute();
+            
+            SignUp signUp = new SignUp().study(Tests.STUDY_ID).password(Tests.PASSWORD);
+            signUp.externalId(externalId);
+            authClient.signUp(signUp).execute();
+            
+            ExternalIdentifierList list = devIdsClient.getExternalIds(null, 50, null, null).execute().body();
```
DwayneJengSage: Consider using idFilter, so we don't have to page through 50 different external IDs to find the one we created.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
```diff
@@ -0,0 +1,124 @@
...
+            for (ExternalIdentifier identifier : list.getItems()) {
+                if (externalId.equals(identifier.getIdentifier())) {
+                    assertTrue(identifier.getAssigned());        
+                }
```
DwayneJengSage: Break statement, so we don't have to keep looping over external IDs after we've found the one we're looking for.

Also, if we don't find the external ID, the test doesn't fail.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
alxdarksage: Updated, also removes the loop below

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdSignUpTest.java
alxdarksage: loop removed

DwayneJengSage (approved):

PR #267, opened 2018-05-21
Tests of new search API.

alxdarksage:
Many of these changes are from the library update, where booleans go from a getFoo accessor to an isFoo() accessor. And one new test class for the search API.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountSummarySearchTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountSummarySearchTest.java
@@ -1,0 +1,226 @@
+package org.sagebionetworks.bridge.sdk.integration;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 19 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountSummarySearchTest.java
```diff
@@ -0,0 +1,226 @@
...
+        if (frenchUser != null) {
+           frenchUser.signOutAndDeleteUser();
+        }
+    }
+    
+    @AfterClass
+    public static void deleteResearcher() throws Exception {
+        if (researcher != null) {
+           researcher.signOutAndDeleteUser();
+        }
```
DwayneJengSage: We create a researcher in several tests, but we only delete one researcher. I think we should create the researcher and worker in the BeforeClass instead. This way, we only need to create them once, and we are more likely to be able to clean them up afterward.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountSummarySearchTest.java
alxdarksage: Moved these to @BeforeClass.

DwayneJengSage (approved):

PR #271, opened 2018-05-29
You can no longer sign in with a phone number using a password unless you've verified the phone number.

alxdarksage:
This is a fix to a test that will no longer work once we tighten security on how accounts are activated and verified. That PR is https://github.com/Sage-Bionetworks/BridgePF/pull/1724

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
@@ -1,6 +1,7 @@
 package org.sagebionetworks.bridge.sdk.integration;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 import org.joda.time.LocalDate;
 import org.junit.After;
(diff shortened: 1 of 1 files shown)


DwayneJengSage (changes requested):
Instead of deleting the last part of the test, should the admin be able to bypass email/phone verification for the purposes of test accounts? Otherwise, there are a lot of things we won't be able to test.

alxdarksage:
As I interpret ParticipantService lines 256 to 273, because it is called from the UserAdminService with shouldSendVerification = false, emailVerified and/or phoneVerified will be set to true if the credential exists and we're not sending verifications. So I believe we already do this. I wouldn't propose making it possible to disable verification through the normal sign up API, just to verify this user has been created.

DwayneJengSage:
I should clarify my comment. This change makes it so that we are no longer able to test intent to participate. Instead of not testing ITP, can we make it so that admins can bypass channel verification, so that we can test ITP?

alxdarksage:
Okay here's another way to do it. We moved the ITP check from sign up to sign in, so it's possible to create an uncontested user, submit an ITP with the same phone number, and then sign in... and the intended effect exists, which is that the user is consented rather than returning a 412.

DwayneJengSage (approved):

PR #274, opened 2018-06-21
Testing changes to app config API

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
```diff
@@ -53,9 +55,10 @@ public void before() throws IOException {
     
     @After
     public void after() throws Exception {
-        AppConfigList list = devApi.getAppConfigs().execute().body();
+        AppConfigList list = devApi.getAppConfigs(true).execute().body();
         for (AppConfig config : list.getItems()) {
-            adminApi.deleteAppConfig(config.getGuid()).execute();
+            assertFalse(config.isDeleted());
+            adminApi.deleteAppConfig(config.getGuid(), true).execute();
```
DwayneJengSage: This test works if the study is empty, but will fail if the study has pre-existing app configs (used for manual testing and local testing). This will also fail if the test fails to physically delete an app config after logically deleting it.

It would be better to keep track of the app configs created by the test and then delete only those.

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
```diff
@@ -158,9 +162,27 @@ public void crudAppConfig() throws Exception {
...
+        int updatedCount = devApi.getAppConfigs(false).execute().body().getItems().size();
+        assertTrue((initialCount+1) == updatedCount);
+        
+        updatedCount = devApi.getAppConfigs(true).execute().body().getItems().size();
+        assertTrue((initialCount+2) == updatedCount);
+        
+        // This should *really* delete the config from the db in order to clean up.
+        adminApi.deleteAppConfig(config.getGuid(), true).execute();
+        int countAfterDelete = devApi.getAppConfigs(true).execute().body().getItems().size();
+        assertTrue(countAfterDelete < updatedCount);
```
DwayneJengSage: Would it be better to get by guid and verify that it throws a 404 not found?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
alxdarksage: Changed this to track the configs made by the test, and only deleting those.

src/test/java/org/sagebionetworks/bridge/sdk/integration/AppConfigTest.java
alxdarksage: Changed to do this.

PR #276, opened 2018-07-05
Need to pause because reauth token is cached for a short period of time.

alxdarksage:
"Fixes" test after adding this caching for concurrent sign ins.


alxdarksage:
Oops. I did this fix twice.

PR #277, opened 2018-07-05
Update survey deletion behavior.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -532,7 +552,7 @@ public void workerCanGetSurveys() throws Exception {
         assertKeysEqual(survey1aKeys, survey1aAgain);
 
         // We only expect the most recently published versions, namely 1b and 2b.
-        SurveyList surveyResourceList = workerApi.getAllPublishedSurveys(IntegTestUtils.STUDY_ID).execute().body();
+        SurveyList surveyResourceList = workerApi.getAllPublishedSurveys(IntegTestUtils.STUDY_ID, false).execute().body();
```
DwayneJengSage: Test for includeDeleted=true?

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -757,6 +781,57 @@ public void displayActionsInAfterRulesValidated() throws Exception {
         }
     }
     
+    @Test
+    public void canRetrieveDeletedSurveys() throws Exception {
+        SurveysApi surveysApi = developer.getClient(SurveysApi.class);
```
DwayneJengSage: What about getPublishedSurveyVersion()? getMostRecentSurveyVersion()?

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -284,8 +284,19 @@ public void getAllVersionsOfASurvey() throws Exception {
...
+        
+        // verify includeDeleted
+        Survey oneVersion = surveyList.getItems().get(0); 
+        surveysApi.deleteSurvey(oneVersion.getGuid(), oneVersion.getCreatedOn(), false).execute();
+        
+        assertEquals(surveyList.getItems().size(),
+                surveysApi.getAllVersionsOfSurvey(key.getGuid(), true).execute().body().getItems().size());
+        
+        assertNotEquals(surveyList.getItems().size(),
+                surveysApi.getAllVersionsOfSurvey(key.getGuid(), false).execute().body().getItems().size());
```
DwayneJengSage: We happen to know that the expected counts here are 2 and 1 respectively. Would it be better to just use thocse constants?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: These should not take an includeDeleted flag; they should resolve to the most recent version or most recently published version that is not (logically) deleted. Removed the flag from the SDK (see new PR, the first PR has been merged) and adjusted the tests here to reflect this. Note the server did not have this flag in its routes or other code for these calls.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -757,6 +786,118 @@ public void displayActionsInAfterRulesValidated() throws Exception {
...
+        surveysApi.publishSurvey(keys2.getGuid(), keys2.getCreatedOn(), false).execute();
+        surveysApi.deleteSurvey(keys2.getGuid(), keys2.getCreatedOn(), false).execute();
+        
+        Thread.sleep(1000);
+        anyDeleted(surveysApi.getPublishedSurveys(true));
+        noneDeleted(surveysApi.getPublishedSurveys(false));
+        
+        try {
+            surveysApi.getPublishedSurveys(true).execute();
+        } catch(EntityNotFoundException e) {
```
DwayneJengSage: Need a fail() here. Otherwise, if this doesn't throw, we'll never know.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: Added here (and one other place in test).

DwayneJengSage (approved):

PR #278, opened 2018-07-09
Updating to use a permanent delete by an admin to clean up test schedule plans.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/SchedulePlanTest.java
```diff
@@ -133,8 +133,14 @@ public void crudSchedulePlan() throws Exception {
         ScheduleList schedules = usersApi.getSchedules().execute().body();
         assertTrue("Schedules exist", !schedules.getItems().isEmpty());
 
-        // Delete
-        newSchedulesApi.deleteSchedulePlan(keys.getGuid()).execute();
+        // Logical delete
+        newSchedulesApi.deleteSchedulePlan(keys.getGuid(), false).execute();
+        
+        SchedulePlan retrieved = newSchedulesApi.getSchedulePlan(keys.getGuid()).execute().body();
+        assertTrue(retrieved.isDeleted());
```
DwayneJengSage: Should we also have a test that we can query deleted schedule plans (developer API and worker API)?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SchedulePlanTest.java
alxdarksage: Added queries of getSchedulePlans() with and without includeDeleted, after logical and physical deletes.

DwayneJengSage (approved):

PR #282, opened 2018-08-07
Simple test of admin sign in APIs.

alxdarksage:
There aren't two set studies so I just test switching to API study, which exercises the code at least and verifies no errors.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/UserManagementTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/UserManagementTest.java
@@ -1,5 +1,6 @@
 package org.sagebionetworks.bridge.sdk.integration;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/UserManagementTest.java
```diff
@@ -81,5 +81,21 @@ public void canCreateAndSignOutAndDeleteUser() throws Exception {
...
+        admin.signOut();
+        
+        ForAdminsApi adminApi = admin.getClient(ForAdminsApi.class);
+        SignIn signIn = new SignIn().study(admin.getStudyId())
+                .email(admin.getEmail()).password((admin.getPassword()));
+        
+        adminApi.adminSignIn(signIn).execute().body();
+        
+        adminApi = admin.getClient(ForAdminsApi.class);
+        adminApi.adminChangeStudy(new SignIn().study("api")).execute().body();
```
DwayneJengSage: Maybe we can leverage the shared study for this. I see 4 uses cases:

1. Admin sign in to api study.
2. Admin sign in to non-api study (shared).
3. Admin switching from api study (to shared study).
4. Admin switching to api study (from shared study).

DwayneJengSage (changes requested):
Looks good. Possible suggestion about improving test coverage.

src/test/java/org/sagebionetworks/bridge/sdk/integration/UserManagementTest.java
alxdarksage: Good point. It slipped my mind that every environment has this other study. I'll change.

DwayneJengSage (approved):

PR #284, opened 2018-08-15
SMS Notifications and Criteria-Managed Topics

DwayneJengSage:
See https://github.com/Sage-Bionetworks/BridgePF/pull/1759

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
@@ -3,8 +3,6 @@ package org.sagebionetworks.bridge.sdk.integration;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
-import java.util.List;
-
 import org.joda.time.LocalDate;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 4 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
```diff
@@ -87,6 +92,15 @@ public void testIntentToParticipate() throws Exception {
             
             ConsentStatus status = session.getConsentStatuses().get(IntegTestUtils.STUDY_ID);
             assertTrue(status.isConsented());
+
+            // This aso creates an SMS notification registration.
```
alxdarksage: typo: "aso" --> "also"

src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
```diff
@@ -87,6 +92,15 @@ public void testIntentToParticipate() throws Exception {
...
+            List<NotificationRegistration> registrationList = user.getClient(ForConsentedUsersApi.class)
+                    .getNotificationRegistrations().execute().body().getItems();
+            assertEquals(1, registrationList.size());
+
+            NotificationRegistration registration = registrationList.get(0);
+            assertEquals(NotificationProtocol.SMS, registration.getProtocol());
+            assertEquals(IntegTestUtils.PHONE.getNumber(), registration.getEndpoint());
```
alxdarksage: It's worth noting that a lot of ITPs might be abandoned before people download and sign in to the application. I think it's fair to send them SMS messages until they opt out via SMS, but it might cost us $$$.

alxdarksage (approved):

src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
DwayneJengSage: As per Server code review, we're no longer registering for subscriptions as part of consent or intent.

src/test/java/org/sagebionetworks/bridge/sdk/integration/IntentToParticipateTest.java
DwayneJengSage: Now moot, since we're deleting this part of the test.

PR #285, opened 2018-08-22
Change test to switch admin between studies.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -126,11 +125,7 @@ public void after() throws Exception {
         
         SurveysApi surveysApi = admin.getClient(SurveysApi.class);
         for (GuidCreatedOnVersionHolder oneSurvey : surveysToDelete) {
-            try {
-                surveysApi.deleteSurvey(oneSurvey.getGuid(), oneSurvey.getCreatedOn(), true).execute();
-            } catch (RuntimeException ex) {
-                LOG.error("Error deleting survey=" + oneSurvey + ": " + ex.getMessage(), ex);
-            }
+            surveysApi.deleteSurvey(oneSurvey.getGuid(), oneSurvey.getCreatedOn(), true).execute();
```
DwayneJengSage: Do we want to delete this try-catch? If we do, then if deleting a survey fails, the test won't attempt to delete any of the other surveys.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -171,6 +166,7 @@ public void testPermanentDeleteWithSharedModule() throws Exception {
         // execute delete
         Exception thrownEx = null;
         try {
+            adminsApi.adminChangeStudy(new SignIn().study("shared")).execute();
             adminSurveysApi.deleteSurvey(retSurvey.getGuid(), retSurvey.getCreatedOn(), true).execute();
```
DwayneJengSage: Do we also need to change the client here as well?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: It's also not consistent with what we're doing elsewhere... restored the try/catch block.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: It is confusing that we have these two clients to do manipulate surveys as an admin... I removed one of them (SurveysApi) and I'm using the ForAdminsApi for both these calls now.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -95,7 +96,8 @@
     private static TestUser worker;
     private static SharedModulesApi sharedDeveloperModulesApi;
     private static SurveysApi sharedSurveysApi;
-    private static SurveysApi adminSurveysApi;
+    //private static SurveysApi adminSurveysApi;
```
DwayneJengSage: Nitpick: Remove commented variable.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -171,15 +173,17 @@ public void testPermanentDeleteWithSharedModule() throws Exception {
...
             fail("expected exception");
         } catch (BadRequestException e) {
             thrownEx = e;
         } finally {
             // finally delete shared module and uploaded schema
             sharedDeveloperModulesApi.deleteMetadataByIdAllVersions(moduleId).execute();
 
-            adminSurveysApi.deleteSurvey(retSurvey.getGuid(), retSurvey.getCreatedOn(), true).execute();
+            admin.getClient(SurveysApi.class).deleteSurvey(retSurvey.getGuid(), retSurvey.getCreatedOn(), true).execute();
+            adminsApi.adminChangeStudy(new SignIn().study("api")).execute();
```
DwayneJengSage: I'm a little confused here. In some places, we call admin.getClient(). In other places, we call the adminsApi directly.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: removed

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: You're right, all these delete APIs are on the ForAdminsApi client, changed all of these to use it.

DwayneJengSage (approved):

PR #291, opened 2018-08-29
Adjusting for the removal of the two synthetic events (adding one back as an automatic custom event).

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ActivityEventTest.java
```diff
@@ -141,18 +141,6 @@ public void automaticCustomEvents() throws Exception {
...
-
-        ActivityEvent twoMonthsBeforeEvent = activityEventMap.get("two_months_before_enrollment");
-        assertNotNull(twoMonthsBeforeEvent);
-        DateTime twoMonthsBeforeTime = twoMonthsBeforeEvent.getTimestamp();
-        Period twoMonthsBeforePeriod = new Period(twoMonthsBeforeTime, enrollmentTime);
-        assertEquals(2, twoMonthsBeforePeriod.getMonths());
-
         // Verify custom event exists and that it's 2 weeks after enrollment
         ActivityEvent twoWeeksAfterEvent = activityEventMap.get("custom:" + TWO_WEEKS_AFTER_KEY);
         assertNotNull(twoWeeksAfterEvent);
```
DwayneJengSage: We'll need to clean up the automatic custom events in the api study (probably manually) and fix up this portion of the test.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ActivityEventTest.java
alxdarksage: Changed the event key.

DwayneJengSage (approved):

PR #296, opened 2018-09-10
Withdraw from study

alxdarksage:
Adding methods to test.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
@@ -2,9 +2,12 @@ package org.sagebionetworks.bridge.sdk.integration;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
```diff
@@ -329,17 +336,13 @@ public void canWithdrawFromAllConsentsInStudy() throws Exception {
             
             assertTrue(RestUtils.isUserConsented(session));
             
-            ParticipantsApi participantsApi = researchUser.getClient(ParticipantsApi.class);
             Withdrawal withdrawal = new Withdrawal().reason("I'm just a test user.");
-            
-            participantsApi.withdrawParticipantFromStudy(session.getId(), withdrawal).execute();
+            testUser.getClient(ForConsentedUsersApi.class).withdrawFromStudy(withdrawal).execute();
```
DwayneJengSage: This test previously tested the researcher withdraw API. Now it tests the user's withdraw API. Was this intentional?

DwayneJengSage: Test for researcher withdrawal was deleted. Both tests have the user withdrawing themselves.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
alxdarksage: Okay, updated the second test to withdraw the user through the participant API, not the user-facing API in both places.

DwayneJengSage (approved):

PR #297, opened 2018-09-10
Researcher Backfill Notifications API

DwayneJengSage:
See https://github.com/Sage-Bionetworks/BridgePF/pull/1775


alxdarksage (approved):
Approved, a shorter method name would be nice if you can come up with one you like.

PR #301, opened 2018-09-21
Logical delete shared modules

alxdarksage:
This is either the last or second-to-last model that needs to be updated to allow developer deletions, so I spent some time on the integration tests trying to verify that we're not logically deleting objects in test cleanup, when we intend to be physically deleting them forever.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleMetadataTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleMetadataTest.java
@@ -109,14 +109,11 @@ public class SharedModuleMetadataTest {
         }
         // also delete created upload schema
         try {
+            adminsApi.deleteAllRevisionsOfUploadSchema(schemaId, true).execute();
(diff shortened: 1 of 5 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/SharedModuleMetadataTest.java
```diff
@@ -574,4 +577,49 @@ public void nonSharedDeveloperCantUpdate() throws Exception {
                 .schemaId(schemaId).schemaRevision(SCHEMA_REV);
         apiDeveloperModulesApi.updateMetadata(moduleId, 1, metadata).execute();
     }
+    
+    @Test
+    public void logicalDelete() throws Exception {
+        try {
```
DwayneJengSage: Test cases for:

* queryAllMetadata w/ includeDeleted=true, includeDeleted=false
* check isDeleted=true for deleted metadata

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #325, opened 2018-11-20
Add integration test for substudy auth checks.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
```diff
@@ -0,0 +1,165 @@
...
+            fail("Should have thrown exception");
+        } catch(EntityNotFoundException e) {
+        }
+    }
+    
+    private void assertAccountHasSubstudies(Set<String> substudies, List<AccountSummary> summaries) {
+        for (AccountSummary summary : summaries) {
+            if (summary.getSubstudyIds().containsAll(substudies)) {
+                return;
+            }
```
DwayneJengSage: If you pass in substudyA, this will return if it finds an account with substudyA+any other study. Since we control the accounts, it might be better to ceck for set equality instead of contains all.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
```diff
@@ -0,0 +1,165 @@
...
+    
+    @After
+    public void deleteSubstudies() {
+        ForAdminsApi adminsApi = admin.getClient(ForAdminsApi.class);
+        for (String substudyId : substudyIdsToDelete) {
+            try {
+                adminsApi.deleteSubstudy(substudyId, true).execute();    
+            } catch(Exception e) {
+                e.printStackTrace();
+            }
```
DwayneJengSage: Does the order of deleting users vs substudies matter? If so, are we sure these two Afters will run in the correct order?

src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
```diff
@@ -0,0 +1,165 @@
...
+    private String createUser(String... substudyIds) throws Exception {
+        String email = IntegTestUtils.makeEmail(SubstudyTest.class);
+        SignUp signUp = new SignUp().email(email).password(Tests.PASSWORD).study(IntegTestUtils.STUDY_ID);
+        signUp.substudyIds(ImmutableList.copyOf(substudyIds));
+                
+        ForAdminsApi adminApi = admin.getClient(ForAdminsApi.class);
+        String userId = adminApi.createUser(signUp).execute().body().getId();
+        userIdsToDelete.add(userId);
+        return userId;
+    }
```
DwayneJengSage: createAdmin() and createUser() are almost identical. Can we consolidate them?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
alxdarksage: This is a weird one. It should matter. If you delete substudies while they are associated to users, you'll get an error, but running the test with the @After methods in either order, this doesn't happen. Maybe they're alphabetized or something.  I combined the methods so whatever is ordering them, if it breaks, it won't break the test.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
alxdarksage: This is a weird one. It should matter. If you delete substudies while they are associated to users, you'll get an error, but running the test with the @after methods in either order, this doesn't happen. Maybe they're alphabetized or something. I combined the methods so whatever is ordering them, if it breaks, it won't break the test.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
alxdarksage: consolidated.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
```diff
@@ -0,0 +1,160 @@
...
+        } catch(EntityNotFoundException e) {
+        }
+    }
+    
+    private void assertAccountHasSubstudies(Set<String> substudies, List<AccountSummary> summaries) {
+        for (AccountSummary summary : summaries) {
+            Set<String> summaryIds = ImmutableSet.copyOf(summary.getSubstudyIds());
+            if (Sets.symmetricDifference(summaryIds, substudies).isEmpty()) {
+                return;
+            }
```
DwayneJengSage: Minor: equals() should do what you want it to do https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#equals(java.lang.Object)

DwayneJengSage (approved):
Looks good. Just one minor note.

src/test/java/org/sagebionetworks/bridge/sdk/integration/SubstudyFilteringTest.java
alxdarksage: Updated to use equals method

DwayneJengSage (approved):

PR #328, opened 2018-12-14
v4 external identifiers api test.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsV3Test.java
```diff
@@ -79,7 +79,7 @@ public void canCRUDExternalIds() throws IOException {
                     .execute().body();
             assertEquals(0, page1.getItems().size());
         } finally {
-            externalIdsClient.deleteExternalIds(identifiers).execute();
+            externalIdsClient.deleteExternalIdsV3(identifiers).execute();
         }
         ExternalIdentifierList page = externalIdsClient.getExternalIds(null, null, prefix, null).execute().body();
```
DwayneJengSage: Should probably use getExternalIdsV3 to ensure coverage of the old APIs. Ideally, all calls in this class should be getExternalIdsV3.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsV4Test.java
```diff
@@ -0,0 +1,219 @@
...
+import org.sagebionetworks.bridge.rest.model.Substudy;
+import org.sagebionetworks.bridge.user.TestUserHelper;
+import org.sagebionetworks.bridge.user.TestUserHelper.TestUser;
+import org.sagebionetworks.bridge.util.IntegTestUtils;
+
+import com.google.common.collect.Lists;
+
+import retrofit2.Response;
+
+public class ExternalIdsV4Test {
```
DwayneJengSage: Should we also have an assignment filter test here?

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsV4Test.java
```diff
@@ -0,0 +1,219 @@
...
+                researcherApi.createExternalId(new ExternalIdentifier().identifier(extIdA)).execute();
+                fail("Should have thrown an exception");
+            } catch (InvalidEntityException e) {
+            }
+
+            // Create a couple of external IDs related to different substudies.
+            ExternalIdentifier extId1 = new ExternalIdentifier().identifier(extIdA).substudyId(idA);
+            ExternalIdentifier extId2 = new ExternalIdentifier().identifier(extIdB).substudyId(idB);
+            researcherApi.createExternalId(extId1).execute();
+            researcherApi.createExternalId(extId2).execute();
```
DwayneJengSage: We create 2 external IDs, but only use 1. Should we be doing something with the other one?

If this is just to test the creation of external IDs, should we get() the external IDs back to verify they were created?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsV4Test.java
alxdarksage: Now that I have added testing of the filtering, the second one is used, so I will probably now leave it.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsV4Test.java
```diff
@@ -0,0 +1,252 @@
...
+
+            // Verify these were both created.
+            ExternalIdentifierList list = researcherApi.getExternalIds(null, null, null, false).execute().body();
```
DwayneJengSage: We should use a prefix for this call, or else we might get random failures when the api study has other external IDs.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ExternalIdsV4Test.java
```diff
@@ -0,0 +1,252 @@
...
+
+            // Cannot create another user with this external ID. This should do nothing and fail quietly.
+            Response<Message> response = researcher.getClient(AuthenticationApi.class).signUp(signUp).execute();
+            assertEquals(201, response.code());
+
+            StudyParticipant participant2 = researcherApi.getParticipantByExternalId(extIdA, false).execute().body();
+            assertEquals(userId, participant2.getId());
+            
+            // verify filtering by assignment since we have assigned one record
+            ExternalIdentifierList all = researcherApi.getExternalIds(null, 50, null, null).execute().body();
```
DwayneJengSage: We should use a prefix in this call (and in the next 2 calls), or we might get random failures when there are other external IDs in the api study.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #329, opened 2019-01-11
Two new survey constraints: YearMonth and PostalCode.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
```diff
@@ -223,6 +228,25 @@ public static Survey getSurvey(Class<?> cls) {
         weightQuestion.setUiHint(UIHint.WEIGHT);
         weightQuestion.setType("SurveyQuestion");
         
+        SurveyQuestion yearMonthQuestion = new SurveyQuestion();
+        YearMonthConstraints c11 = new YearMonthConstraints();
+        c11.setDataType(DataType.YEARMONTH);
```
DwayneJengSage: Should we add earliest, latest, and allowFuture here, just to get coverage on all attributes?

src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
```diff
@@ -238,6 +262,8 @@ public static Survey getSurvey(Class<?> cls) {
         elements.add(bloodpressureQuestion);
         elements.add(heightQuestion);
         elements.add(weightQuestion);
+        elements.add(yearMonthQuestion);
+        elements.add(postalCodeQuestion);
```
DwayneJengSage: Is there a test somewhere where we receive the survey back from the server to verify round-trip survey marshalling?  (Including all the new attributes: allowFuture, earliest, latest, country code, and sparse zip codes?)

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
alxdarksage: I added a test like this because it doesn't exist, moved one test into it, and found a couple of bugs in earlier mappings for earlier constraints that were added. I fixed those and bumped the version of the SDK, and now I am importing that version (the test also now includes a test of these new constraint types).

src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
alxdarksage: Yes, added and then added to test, described below. That test in turn led me to add a few more of these fields to the TestSurvey that were previously left null.

src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
```diff
@@ -29,10 +31,34 @@
...
+    public static final Integer DURATION_QUESTION_MIN_VALUE = 1;
+    public static final Integer INT_QUESTION_MAX_VALUE = 8;
+    public static final Integer INT_QUESTION_MIN_VALUE = 0;
+    public static final Integer INT_QUESTION_STEP = 1;
+    public static final Unit INT_QUESTION_UNIT = Unit.DAYS;
+    public static final BigDecimal DECIMAL_QUESTION_STEP = BigDecimal.valueOf(0.1d);
+    public static final BigDecimal DECIMAL_QUESTION_MAX_VALUE = BigDecimal.valueOf(10.0d);
+    public static final BigDecimal DECIMAL_QUESTION_MIN_VALUE = BigDecimal.valueOf(0.0d);
+    public static final DateTime DATETIME_LATEST_VALUE = DateTime.parse("2020-12-31").withZone(DateTimeZone.UTC);
+    public static final DateTime DATETIME_EARLIEST_VALUE = DateTime.parse("2000-01-01").withZone(DateTimeZone.UTC);
```
DwayneJengSage: Should we test that this works with a timezone other than UTC? (If it doesn't work, we should file a JIRA to figure out why as a separate code change.)

DwayneJengSage (approved):
Looks good. Just one comment about timezones, not a blocker for this pull request.

PR #332, opened 2019-02-06
Adding check that the substudy membership is being recorded in the health data record.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
```diff
@@ -38,34 +40,54 @@
...
+            VersionHolder version = admin.getClient(ForAdminsApi.class).createSubstudy(substudy).execute().body();
+            substudy.setVersion(version.getVersion());
+        }
+        
         // developer is to ensure schemas exist. user is to do uploads
         worker = TestUserHelper.createAndSignInUser(UploadTest.class, false, Role.WORKER);
         developer = TestUserHelper.createAndSignInUser(UploadTest.class, false, Role.DEVELOPER);
-        user = TestUserHelper.createAndSignInUser(UploadTest.class, true);
-        admin = TestUserHelper.getSignedInAdmin();
+        researcher = TestUserHelper.createAndSignInUser(UploadTest.class, false, Role.RESEARCHER);
```
DwayneJengSage: What is the researcher used for?

src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
```diff
@@ -271,6 +300,8 @@ public void miscTests() throws Exception {
         // Validate the record data.
         HealthDataRecord record = status.getRecord();
         assertEquals(SharingScope.ALL_QUALIFIED_RESEARCHERS, record.getUserSharingScope());
+        assertEquals(1, record.getUserSubstudyMemberships().size());
+        assertEquals("", record.getUserSubstudyMemberships().get(SUBSTUDY_ID));
```
DwayneJengSage: Would it be difficult to assign an external ID to the user, so we can also verify the external ID is saved to the map?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
alxdarksage: Nothing at this point and could be removed, but given the comment below to add an external ID, we will need a researcher for that.

src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
alxdarksage: I've added this and we verify the association to a subsidy is carried over into the health data record.

DwayneJengSage (approved):

PR #343, opened 2019-04-01
Tests of substudy partitioning of reports.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -424,7 +459,76 @@ public void userCanCRUDSelfReports() throws Exception {
...
+        } catch(UnauthorizedException e) {
+            // expected, and from the correct call.
+        }
+        
+        // Make it public, and the user *can* see it
+        index.setPublic(true);
+        devReportClient.updateStudyReportIndex(index.getIdentifier(), index).execute();
+        
+        ReportDataList list = reportsApi.getStudyReportRecords(reportId, SEARCH_START_DATE, SEARCH_END_DATE).execute().body();
+        assertEquals(2, list.getItems().size());
```
DwayneJengSage: Instead of just counting the reports, we should also validate the content (most notably the substudyIDs).

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -424,7 +459,76 @@ public void userCanCRUDSelfReports() throws Exception {
...
+        substudyScopedUser = new TestUserHelper.Builder(ReportTest.class).withConsentUser(true)
+                .withSubstudyIds(ImmutableSet.of(substudy2.getId())).createAndSignInUser();
+        
+        String healthCode = worker.getClient(ParticipantsApi.class)
+                .getParticipantById(substudyScopedUser.getUserId(), false).execute().body().getHealthCode();
+
+        ForWorkersApi workerApi = worker.getClient(ForWorkersApi.class);
+        ReportDataForWorker data1 = makeReportDataForWorker(healthCode, DATE1, "asdf", "A");
+        data1.setHealthCode(healthCode);
+        data1.setSubstudyIds(ImmutableList.of(substudy1.getId()));
```
DwayneJengSage: We should also check the workers can get reports back with substudies.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -424,7 +459,76 @@ public void userCanCRUDSelfReports() throws Exception {
...
+        ReportData data2 = makeReportData(DATE2, "asdf", "B");
+        devReportClient.addStudyReportRecord(reportId, data1).execute();
+        devReportClient.addStudyReportRecord(reportId, data2).execute();
+        
+        // Not a member of the substudy used for these report records
+        substudyScopedUser = new TestUserHelper.Builder(ReportTest.class).withConsentUser(true)
+                .withSubstudyIds(ImmutableSet.of(substudy2.getId())).createAndSignInUser();
+        StudyReportsApi reportsApi = substudyScopedUser.getClient(StudyReportsApi.class);
+        ReportIndex index = reportsApi.getStudyReportIndex(reportId).execute().body();
```
DwayneJengSage: Should there be an equivalent test for getParticipantReportIndex()?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: Once set with the first report, substudy IDs are only persisted on the index, for all the records, so that field is null when you retrieve the reports. I did add tests of all of this.

I guess we could set those as read-only values in the report data records as well when reading from the server, but it didn't come up as something useful when implementing UI to create and edit these.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #345, opened 2019-04-05
Survey Schema new format

DwayneJengSage:
Integ tests for https://github.com/Sage-Bionetworks/BridgePF/pull/1851

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/HealthDataTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/HealthDataTest.java
@@ -10,6 +10,7 @@ import static org.junit.Assert.fail;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import org.apache.commons.lang3.RandomStringUtils;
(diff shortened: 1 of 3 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/HealthDataTest.java
```diff
@@ -258,8 +280,13 @@ public void submitBySurvey() throws Exception {
...
 
         Map<String, String> returnedDataMap = RestUtils.toType(record.getData(), Map.class);
-        assertEquals(1, returnedDataMap.size());
+        assertEquals(2, returnedDataMap.size());
         assertEquals("C", returnedDataMap.get("answer-me"));
+
+        // For some bizarre reason, GSON won't parse the data.get("answers"). Use Jackson to parse it into a JsonNode.
+        JsonNode rawAnswersNode = DefaultObjectMapper.INSTANCE.readTree(returnedDataMap.get("answers"));
+        assertEquals(1, rawAnswersNode.size());
+        assertEquals("C", rawAnswersNode.get("answer-me").textValue());
```
alxdarksage: It's a little worrisome that GSON doesn't work here, given that the Android SDK uses GSON and this could potentially be confusing down the road. Maybe we should add this to the API documentation for the getHealthDataByCreatedOn() and similar calls. that return a HealthDataRecord (there are not many).

alxdarksage (approved):
One comment about this bug you found in GSON possibly confusing our own end users.

PR #346, opened 2019-04-05
integ tests for survey ID uniqueness

DwayneJengSage:
Integ tests for https://github.com/Sage-Bionetworks/BridgePF/pull/1852


alxdarksage:
I filed BRIDGE-2428 to capture an eventual phase out of GUID for the identifier, at least as documented for new users of the system.

alxdarksage (approved):

PR #350, opened 2019-04-17
Adding some tests to cover functionality that was being tested in BridgePF by starting full Spring context

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -197,6 +212,89 @@ private void dailyTaskAt4Times() throws IOException {
...
         schedulePlanGuidList.add(planGuid);
     }
+    
+    private void compoundDailyTask() throws IOException {
+        UploadFieldDefinition fieldDef = new UploadFieldDefinition();
+        fieldDef.setName("field");
+        fieldDef.setType(UploadFieldType.STRING);
+        UploadSchema schema = new UploadSchema();
+        schema.setName("Schema");
+        schema.setSchemaId("schemaId"+RandomStringUtils.randomAlphabetic(4));
```
DwayneJengSage: We should use Tests.randomIdentifier(), so we tag the schema ID with the test that created it.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -197,6 +212,89 @@ private void dailyTaskAt4Times() throws IOException {
...
+        schema.setSchemaType(UploadSchemaType.IOS_DATA);
+        schema.setFieldDefinitions(Lists.newArrayList(fieldDef));
+        
+        // create it
+        UploadSchemasApi schemasApi = developer.getClient(UploadSchemasApi.class);
+        schemaKeys = schemasApi.createOrUpdateUploadSchema(schema).execute().body();
```
DwayneJengSage: Minor: createUploadSchema() is preferred over createOrUpdateUploadSchema().

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -230,11 +328,19 @@ private void miniStudyBurstSchedule() throws Exception {
...
+        }
         try {
             SchedulesApi schedulesApi = admin.getClient(SchedulesApi.class);
             for (String oneSchedulePlanGuid : schedulePlanGuidList) {
                 schedulesApi.deleteSchedulePlan(oneSchedulePlanGuid, true).execute();
             }
+            if (surveyKeys != null) {
+                admin.getClient(SurveysApi.class)
+                        .deleteSurvey(surveyKeys.getGuid(), surveyKeys.getCreatedOn(), true).execute();
+            }
```
DwayneJengSage: Is there any reason why this delete is inside the try, but the other delete is outside?

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -508,7 +614,226 @@ public void scheduleWithMultipleEventIds() throws Exception {
...
+            act.setStartedOn(DateTime.now());
+            act.setFinishedOn(DateTime.now());
+        }
+        userApi.updateScheduledActivities(actList.getItems()).execute();
+        
+        Set<String> guids = new HashSet<>();
+        
+        ForwardCursorScheduledActivityList list = userApi.getCompoundActivityHistory(
+                taskId, startsOn, endsOn, null, 10).execute().body();
+        assertEquals(10, list.getItems().size());
```
DwayneJengSage: There may be other scheduled activities using that task ID. Should we create a unique activity label and filter on that label?

Similarly for getTaskHistory().

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -508,7 +614,226 @@ public void scheduleWithMultipleEventIds() throws Exception {
...
+                "not the right ID", startsOn, endsOn, null, null).execute().body();
+        assertTrue(list2.getItems().isEmpty());
+        
+        ForwardCursorScheduledActivityList list3 = userApi.getSurveyHistory(
+                taskId, startsOn, endsOn, null, null).execute().body();
+        assertTrue(list3.getItems().isEmpty());
+
+        ForwardCursorScheduledActivityList list4 = userApi.getTaskHistory(
+                taskId, startsOn, endsOn, null, null).execute().body();
+        assertTrue(list4.getItems().isEmpty());
```
DwayneJengSage: Similarly, this may pick up other tasks with the same task ID.

Similarly for getTaskHistory().

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -508,7 +614,226 @@ public void scheduleWithMultipleEventIds() throws Exception {
...
+        
+        dailyTaskAt4Times();
+        
+        ForConsentedUsersApi userApi = user.getClient(ForConsentedUsersApi.class);
+        
+        String activityGuid = null;
+        
+        ScheduledActivityListV4 actList = userApi.getScheduledActivitiesByDateRange(startsOn, endsOn).execute().body();
+        for (ScheduledActivity act : actList.getItems()) {
+            activityGuid = act.getGuid().split(":")[0];
```
DwayneJengSage: Are we sure this is what we want? This basically uses the activityGuid from the last activity in the list, which might not be the one we're looking for. We should set a unique activity label and filter on activity label.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -508,7 +614,226 @@ public void scheduleWithMultipleEventIds() throws Exception {
...
+        }
+        userApi.updateScheduledActivities(actList.getItems()).execute();
+        
+        Set<String> guids = new HashSet<>();
+        
+        ForwardCursorScheduledActivityList list = userApi.getActivityHistory(
+                activityGuid, startsOn, endsOn, null, 10).execute().body();
+        assertEquals(10, list.getItems().size());
+        for (ScheduledActivity act : list.getItems()) {
+            guids.add(act.getGuid());
```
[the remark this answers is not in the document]
alxdarksage: Added.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: Changed

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: Changed

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: No, the admin isn't deleted, it can come after the finally block. Moved.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: All the scheduled activities are created for a user that is created for this specific test, so they should never conflict even if other users have scheduled activities with the same task ID.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: See above

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: I rewrote all of these to filter out tasks with the wrong label (using runId) to avoid this problem. Verified by leaving all the schedule plans from one test and re-running the test suite again.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: We now filter before retrieving this value from the list, so it will be correct.

alxdarksage: Now this is done after filtering on the label, so it is the GUID that we want.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -508,7 +601,215 @@ public void scheduleWithMultipleEventIds() throws Exception {
...
+        
+        ForwardCursorScheduledActivityList list = filterPages(true, (offsetKey) ->
+            userApi.getTaskHistory(TASK_ID, STARTS_ON, ENDS_ON, offsetKey, 10).execute().body());
+        
+        for (ScheduledActivity act : list.getItems()) {
+            assertEquals(TASK_ID, act.getActivity().getTask().getIdentifier());
+            assertEquals(EST, act.getScheduledOn().getZone());
+        }
+        
+        assertEquals(16, list.getItems().size());
```
DwayneJengSage: We should check request params.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -246,43 +349,31 @@ public void after() throws Exception {
...
-        usersApi.getScheduledActivitiesByDateRange(startTime, endTime).execute();
+        usersApi.getScheduledActivitiesByDateRange(STARTS_ON, ENDS_ON).execute();
 
         // getTaskHistory() uses a secondary global index. Sleep for 2 seconds to help make sure the index is consistent.
         Thread.sleep(2000);
 
         // Now we should see those in the latest API:
-        ForwardCursorScheduledActivityList list = usersApi.getTaskHistory("task:AAA", startTime, endTime, null, 10)
-                .execute().body();
+        ForwardCursorScheduledActivityList list = filterPages(false, (offsetKey) ->
```
DwayneJengSage: Shouldn't this be true, since we expect at least 2 pages?

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
```diff
@@ -508,7 +601,215 @@ public void scheduleWithMultipleEventIds() throws Exception {
...
+            throws Exception {
+        ScheduledActivityListV4 filteredList = new ScheduledActivityListV4();
+        
+        List<ScheduledActivity> activities = new ArrayList<>();
+        ScheduledActivityListV4 results = caller.apply(null);
+        for (ScheduledActivity act : results.getItems()) {
+            if (act.getActivity().getLabel().equals(runId)) {
+                activities.add(act);    
+            }
+        }
```
DwayneJengSage: Should this have pagination logic?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: Yes, changed to true

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
alxdarksage: This API is not paged. We limit the date range to 32 days but that's it.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityTest.java
DwayneJengSage: This should be true, since we expect at least 2 pages.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #352, opened 2019-04-24
Potential fix for a time conflict with the enrollment date cutting off one day of results.

alxdarksage:


DwayneJengSage:
As per IM conversation, I believe my theory yesterday was wrong.

PR #361, opened 2019-05-31
Porting a couple of tests of scheduling from BridgePF to the integration tests.

alxdarksage:
They are a little different because we can't fix time on the server, but demonstrate that time zones are handled in a consistent fashion.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduleActivityOnceTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduleActivityOnceTest.java
@@ -1,0 +1,115 @@
+package org.sagebionetworks.bridge.sdk.integration;
+
+import static org.junit.Assert.assertEquals;
+import static org.sagebionetworks.bridge.rest.model.Role.DEVELOPER;
+
+import com.google.common.collect.ImmutableList;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import org.sagebionetworks.bridge.rest.api.ForAdminsApi;
+import org.sagebionetworks.bridge.rest.api.ForConsentedUsersApi;
+import org.sagebionetworks.bridge.rest.api.SchedulesApi;
+import org.sagebionetworks.bridge.rest.api.StudiesApi;
+import org.sagebionetworks.bridge.rest.model.Activity;
+import org.sagebionetworks.bridge.rest.model.ActivityType;
+import org.sagebionetworks.bridge.rest.model.Criteria;
+import org.sagebionetworks.bridge.rest.model.CriteriaScheduleStrategy;
+import org.sagebionetworks.bridge.rest.model.GuidVersionHolder;
+import org.sagebionetworks.bridge.rest.model.Schedule;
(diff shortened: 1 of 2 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityRecurringTest.java
```diff
@@ -0,0 +1,181 @@
...
+        String ytz2 = now.withZone(YTZ).plusDays(1).toLocalDate().toString()+Y_TIME_OF_DAY;
+        String ytz3 = now.withZone(YTZ).plusDays(2).toLocalDate().toString()+Y_TIME_OF_DAY;
+        String ytz4 = now.withZone(YTZ).plusDays(3).toLocalDate().toString()+Y_TIME_OF_DAY;
+        
+        // Get three tasks in the Gilbert Island for today and next 2 days
+        ScheduledActivityList activities = usersApi.getScheduledActivities("+12:00", 2, null).execute().body();
+        assertEquals(3, activities.getItems().size());
+        assertEquals(mtz1, activities.getItems().get(0).getScheduledOn().toString());
+        assertEquals(mtz2, activities.getItems().get(1).getScheduledOn().toString());
+        assertEquals(mtz3, activities.getItems().get(2).getScheduledOn().toString());
```
DwayneJengSage: This test assumes there are no other schedules in the study, which might not be the case. It would be better to use a label and then filter on that label.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduleActivityOnceTest.java
```diff
@@ -0,0 +1,113 @@
...
+        admin = TestUserHelper.getSignedInAdmin();
+        developer = TestUserHelper.createAndSignInUser(ScheduleActivityOnceTest.class, true, DEVELOPER);
+        SignUp signUp = new SignUp().dataGroups(ImmutableList.of("sdk-int-1"));
+        user = new TestUserHelper.Builder(ScheduleActivityOnceTest.class).withConsentUser(true).withSignUp(signUp)
+                .createAndSignInUser();
+        
+        SchedulePlanList list = developer.getClient(SchedulesApi.class).getSchedulePlans(true).execute().body();
+        for (SchedulePlan plan : list.getItems()) {
+            admin.getClient(ForAdminsApi.class).deleteSchedulePlan(plan.getGuid(), true).execute();
+        }
```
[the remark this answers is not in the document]
alxdarksage: Switched to the use of a label and filtering

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityRecurringTest.java
alxdarksage: Filtering on schedule guid

DwayneJengSage:
Should there also be an update to ScheduledActivityRecurringTest?

alxdarksage:
Updated to include activity filtering in ScheduledActivityRecurringTest test.

DwayneJengSage (approved):

PR #365, opened 2019-06-25
Templates test

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/TemplateTest.java
```diff
@@ -0,0 +1,144 @@
...
+        
+        devsApi.deleteTemplate(template.getGuid(), false).execute();
+        
+        retrieved = devsApi.getTemplate(template.getGuid()).execute().body();
+        assertTrue(retrieved.isDeleted());
+        
+        // get a page of templates with logical deletes
+        TemplateList list = devsApi.getTemplates(EMAIL_ACCOUNT_EXISTS.name(), null, null, false).execute()
+                .body();
+        assertTrue(list.getItems().isEmpty());
```
DwayneJengSage: Long-term, this might come back to bite us. When we eventually wire up these templates to the actual emails, and we have integ tests that test these scenarios, we may end up needing these template to exist and be not deleted.

It would be better if you check the list and make sure none of them match the expected template guid.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/TemplateTest.java
alxdarksage: Very true. Changed both places (find and don't find in results).

DwayneJengSage (approved):

PR #370, opened 2019-07-23
Reauth enabled for new studies by default

DwayneJengSage:
See https://github.com/Sage-Bionetworks/BridgeServer2/pull/69 and https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/570


alxdarksage (changes requested):
Verify that setting field to null on update = setting it to Boolean.FALSE (or throws validation exception).

alxdarksage (approved):

PR #390, opened 2019-09-26
Master Scheduler Test

janaechan:


src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.sagebionetworks.bridge.rest.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.rest.model.DateTimeHolder;
+import org.sagebionetworks.bridge.rest.model.MasterSchedulerConfig;
+import org.sagebionetworks.bridge.rest.model.Message;
+import org.sagebionetworks.bridge.user.TestUserHelper;
+import org.sagebionetworks.bridge.user.TestUserHelper.TestUser;
+
+import retrofit2.Response;
+
+public class MasterSchedulerTest {
```
DwayneJengSage: We also need test cases for:

* successful get()
* list(), making sure that it includes the config you created.

Also, a lot of these tests have the same setup (creating the config). It might be easier to combine them all into a single test instead.

src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,116 @@
...
+        
+        Response<Message> response = adminApi.deleteSchedulerConfig(SCHEDULE_ID).execute();
+        assertEquals(200, response.code());
+    }
+    
+    @Test
+    public void testGetSchedulerStatus() throws Exception {
+        DateTimeHolder dateTime = adminApi.getSchedulerStatus().execute().body();
+        
+        assertNull(dateTime.getDateTime());
```
DwayneJengSage: The scheduler may or may not have run. I think the best way to test it is to test that the DateTimeHolder exists, but not necessarily verify its contents.

src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
```diff
@@ -323,4 +326,19 @@ public static void deleteExternalId(ExternalIdentifier externalId) throws Except
...
     }    
+    
+    public static MasterSchedulerConfig getMastSchedulerConfig() {
+        ObjectNode objNode = JsonNodeFactory.instance.objectNode();
+        objNode.put("a", true);
+        objNode.put("b", "string");
+        
+        MasterSchedulerConfig config = new MasterSchedulerConfig();
+        config.setScheduleId("test-schedule-id");
+        config.setCronSchedule("testCronSchedule");
```
DwayneJengSage: We should use a real cron expression, so that if the config somehow gets left behind in the DDB table, it won't cause the scheduler to start throwing errors. (Maybe "0 0 0 1 * ?", which runs at midnight on the 1st of every month, so that the Master Scheduler won't actually try to execute it.)

Same with the update test.

src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
```diff
@@ -323,4 +326,19 @@ public static void deleteExternalId(ExternalIdentifier externalId) throws Except
...
+        config.setRequestTemplate(objNode);
+        config.setSqsQueueUrl("testSysQueueUrl");
```
DwayneJengSage: Let's point this at the dummy queue (https://sqs.us-east-1.amazonaws.com/649232250620/Bridge-UDD-Dummy-Queue), so that if it gets executed, it never actually gets sent anywhere.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    public void testMasterSchedulerConfig() throws IOException {
+        try {
+            adminApi.getSchedulerConfig(SCHEDULE_ID).execute().body();
+            fail("expected exception");
+        } catch (EntityNotFoundException e) {
+            assertEquals("MasterSchedulerConfig not found.", e.getMessage());
+        }
+        
+        MasterSchedulerConfig result = adminApi.createSchedulerConfig(config).execute().body();
+        assertNotNull(result);
```
DwayneJengSage: We should verify the properties of the config we created. Similarly for get and list.

src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertEquals(updatedResult.getCronSchedule(), UPDATED_CRON);
+        assertEquals(updatedResult.getVersion(), new Long(2));
+        
+        MasterSchedulerConfigList updatedConfigList = adminApi.getAllSchedulerConfigs().execute().body();
+        List<MasterSchedulerConfig> updatedConfigs = updatedConfigList.getItems();
+        assertNotEquals(configs, updatedConfigs);
+        
+        try {
+            adminApi.deleteSchedulerConfig("delete-schedule-id").execute().body();
+        } catch (EntityNotFoundException e) {
```
DwayneJengSage: We should put a fail() in the try, so if the delete call doesn't throw, we know about it.

src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,115 @@
...
+            assertEquals("MasterSchedulerConfig not found.", e.getMessage());
+        }
+        
+        Response<Message> response = adminApi.deleteSchedulerConfig(SCHEDULE_ID).execute();
+        assertEquals(200, response.code());
```
DwayneJengSage: Perhaps verify with a get() that the config no longer exists.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,149 @@
...
+        assertEquals(result.getVersion(), config.getVersion());
+        
+        returnedRequestTemplateMap = mapper.convertValue(result.getRequestTemplate(), Map.class);
+        assertEquals(2, returnedRequestTemplateMap.size());
+        assertEquals("string", returnedRequestTemplateMap.get("a"));
+        assertEquals(true, returnedRequestTemplateMap.get("b"));
+        
+        MasterSchedulerConfigList configList = adminApi.getAllSchedulerConfigs().execute().body();
+        List<MasterSchedulerConfig> configs = configList.getItems();
+        assertEquals(1, configs.size());
```
DwayneJengSage: We currently have multiple scheduler configs in all environments. And as we add more, we don't want to have to update this test. Instead of asserting on the size of the list, we should walk through the list until we find the schedule ID we're looking for.

src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,149 @@
...
+        } catch (EntityNotFoundException e) {
+            assertEquals("MasterSchedulerConfig not found.", e.getMessage());
+        }
+        
+        Response<Message> response = adminApi.deleteSchedulerConfig(SCHEDULE_ID).execute();
+        assertEquals(200, response.code());
+        
+        try {
+            adminApi.getSchedulerConfig(SCHEDULE_ID).execute().body();
+        } catch (EntityNotFoundException e) {
```
DwayneJengSage: We also need a fail here.

src/test/java/org/sagebionetworks/bridge/sdk/integration/MasterSchedulerTest.java
```diff
@@ -0,0 +1,149 @@
...
+            assertEquals("MasterSchedulerConfig not found.", e.getMessage());
+        }
+        
+        configList = adminApi.getAllSchedulerConfigs().execute().body();
+        assertEquals(0, configList.getItems().size());
```
DwayneJengSage: Similarly, we have other scheduler configs in all environments. Instead of verifying that we have no configs, we should verify that our schedule ID isn't in the list.

DwayneJengSage (changes requested):

DwayneJengSage:
Looks good. Looks like JavaSDK is now on 0.19.13, so you'll need to bump it to 0.19.14.

DwayneJengSage (approved):

PR #391, opened 2019-09-27
Adding test for year type survey question.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -196,64 +234,66 @@ public void canRountripSurvey() throws Exception {
...
         assertEquals("BooleanConstraints", booleanConstraints.getType());
-        assertEquals(DataType.BOOLEAN, booleanConstraints.getDataType());
+        assertEquals(BOOLEAN, booleanConstraints.getDataType());
 
         // Date question
         SurveyQuestion dateQuestion = getQuestion(survey, DATE_ID);
         DateConstraints dateConstraints = (DateConstraints)dateQuestion.getConstraints();
-        assertEquals(DataType.DATE, dateConstraints.getDataType());
+        assertEquals(DATE, dateConstraints.getDataType());
+        assertTrue(dateConstraints.isAllowFuture()); // true by default
```
DwayneJengSage: Nitpick: allowFuture is false by default. This field is true because in TestSurvey, we set it to true. (Here and in a few other places in this test.)

src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
```diff
@@ -282,6 +284,18 @@ public static Survey getSurvey(Class<?> cls) throws Exception {
         postalCodeQuestion.setIdentifier(POSTALCODE_ID);
         postalCodeQuestion.setUiHint(UIHint.POSTALCODE);
         Tests.setVariableValueInObject(postalCodeQuestion, "type", "SurveyQuestion");
+        
+        SurveyQuestion yearQuestion = new SurveyQuestion();
+        YearConstraints yc = new YearConstraints();
+        yc.setAllowFuture(true);
+        yc.setAllowPast(false);
+        yc.setLatestValue("2020");
```
DwayneJengSage: For completeness, should we also set the EarliestValue?

src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
```diff
@@ -282,6 +284,18 @@ public static Survey getSurvey(Class<?> cls) throws Exception {
...
+        yc.setDataType(DataType.YEAR);
+        yearQuestion.setConstraints(yc);
+        yearQuestion.setPrompt("When is your next MRI?");
+        yearQuestion.setIdentifier(YEAR_ID);
+        yearQuestion.setUiHint(UIHint.SLIDER);
```
DwayneJengSage: Similarly, for completeness, should we also test our new UIHint.YEAR?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
alxdarksage: I've consistently changed the time-based constraints to flip the defaults, and then test that they are flipped.

src/test/java/org/sagebionetworks/bridge/sdk/integration/TestSurvey.java
alxdarksage: Added

DwayneJengSage (approved):

PR #394, opened 2019-10-18
test use of identifier prefix to access surveys

janaechan:


DwayneJengSage (changes requested):
Needs the identifier equivalent of getMostRecentSurveyVersionAndDelete(). Other than that, looks good.

DwayneJengSage (approved):

PR #395, opened 2019-10-30
Checking for synapse user ID

alxdarksage:


DwayneJengSage (commented):
What about the new get user by synapse ID APIs?

DwayneJengSage (changes requested):
What about the new get user by synapse ID APIs?

alxdarksage:
Added tests for the two getParticipantBySynapseUserId() methods (one for worker, one for researcher). Note that I also had to change BridgeIntegTestUtils and bump its version.

DwayneJengSage (approved):

PR #403, opened 2019-11-22
Very marking survey options as exclusive works

alxdarksage:
The survey engine will deselect all other options when an exclusive option is checked, and probably disable them as well. It is used for a "none of the above" option which is not the same as an "other" option.


src/test/java/org/sagebionetworks/bridge/sdk/integration/SurveyTest.java
```diff
@@ -311,7 +311,7 @@ public void canRountripSurvey() throws Exception {
         assertTrue(multiValueConstraints.isRequired());        
         assertEquals("MultiValueConstraints", multiValueConstraints.getType());
         List<SurveyQuestionOption> options = multiValueConstraints.getEnumeration();
-        assertEquals(5, options.size());
+        assertEquals(6, options.size());
         SurveyQuestionOption option = options.get(0);
```
DwayneJengSage: Do we want to change the below to option 5 and check that the isExclusive flag is true?

DwayneJengSage (approved):

PR #404, opened 2019-11-25
Introduce superadmin role

alxdarksage:
Note to self: I merged and then had to revert changes to BridgeIntegTestUtils that will need to be redone.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/SelfStudyTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/SelfStudyTest.java
@@ -1,12 +1,16 @@
 package org.sagebionetworks.bridge.sdk.integration;
 
 import static org.junit.Assert.assertEquals;
+import static org.sagebionetworks.bridge.rest.model.Role.ADMIN;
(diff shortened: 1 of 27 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/SelfStudyTest.java
```diff
@@ -48,19 +47,17 @@ public void getCurrentStudy() throws Exception {
         assertEquals("api", study.getIdentifier());
     }
     
-    @Test(expected = UnauthorizedException.class)
```
DwayneJengSage: It looks like you accidentally deleted the test annotation from this test and the next.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/SelfStudyTest.java
alxdarksage: I added the researcher expected exception (this would have failed after merging) and added a test to verify study admins can update their own studies as admins.

DwayneJengSage (approved):

PR #407, opened 2019-12-12
These tests were not doing what they appeared to be doing

alxdarksage:
We needed to separate the scoped developer out because in some tests, the scoped developer was being used to manipulate reports from a user in a different substudy. This was only working because of a server bug where this scoped developer could retrieve the account... this test has been fixed to reflect the fix in the server code.


DwayneJengSage (commented):
Looks good to me.

PR #410, opened 2020-01-15
Rollback in server fix reverts this test to its original assertions.

alxdarksage:


DwayneJengSage:
Similarly, I'd prefer this to be a 1-click revert of https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/408

PR #417, opened 2020-03-19
Tag API tests.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/TagTest.java
```diff
@@ -0,0 +1,50 @@
...
+        tagsApi.deleteTag("cat2:tag2.1").execute();
+        tagsApi.deleteTag("cat2:tag2.2").execute();
+        tagsApi.deleteTag("tag3").execute();
+        
+        allTags = tagsApi.getTags().execute().body();
+        assertFalse(allTags.containsKey("default") && allTags.get("default").contains("tag3"));
+        assertFalse(allTags.containsKey("cat1") && allTags.get("cat1").contains("tag1.1"));
+        assertFalse(allTags.containsKey("cat1") && allTags.get("cat1").contains("tag1.2"));
+        assertFalse(allTags.containsKey("cat2") && allTags.get("cat2").contains("tag2.1"));
+        assertFalse(allTags.containsKey("cat2") && allTags.get("cat2").contains("tag2.2"));
```
DwayneJengSage: You should break up these assertFalse() statements. If the first clause is false, then the second clause will never be checked.

src/test/java/org/sagebionetworks/bridge/sdk/integration/TagTest.java
```diff
@@ -0,0 +1,50 @@
...
+        assertEquals(allTags.get("cat2"), ImmutableList.of("tag2.1", "tag2.2"));
+        
+        tagsApi.deleteTag("cat1:tag1.1").execute();
+        tagsApi.deleteTag("cat1:tag1.2").execute();
+        tagsApi.deleteTag("cat2:tag2.1").execute();
+        tagsApi.deleteTag("cat2:tag2.2").execute();
+        tagsApi.deleteTag("tag3").execute();
+        
+        allTags = tagsApi.getTags().execute().body();
+        assertFalse(allTags.containsKey("default") && allTags.get("default").contains("tag3"));
```
DwayneJengSage: There might be other default tags in the api study. It would be better to merely check that your specific tag3 doesn't exist, rather than checking that there are no default tags.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/TagTest.java
alxdarksage: The first part was just to prevent a null pointer exception before the second test... if the category doesn't exist, the specific tag in that category can't exist so it's okay if only the first part is evaluated... I can rewrite this to be clearer.

DwayneJengSage (approved):

PR #424, opened 2020-05-01
Switching the integration tests from studies to apps

alxdarksage:
The IntegTestUtils package will need to be merged first, in addition to the SDK.


src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityRecurringTest.java
```diff
@@ -143,19 +142,19 @@ public void retrievalActivitiesAcrossTimeAndTimeZones() throws Exception {
...
-        // One activity was created beyond the window, in +12:00 land.... that is not returned because although it 
-        // exists, we filter it out from the persisted activities retrieved from the db.
+        // Cross the dateline to the prior day. You get 3 activities (yesterday, today, and tomorrow). Two activities 
+        // were created beyond the window, in +12:00 land.... those are not returned because although it exists, we 
+        // filter it out from the persisted activities retrieved from the db.
         activities = filterList(usersApi.getScheduledActivities("-12:00", 2, null).execute().body(),
                 schedulePlan.getGuid());
-        assertEquals(4, activities.getItems().size());
+        
+        assertEquals(3, activities.getItems().size());
```
DwayneJengSage, line 151: Is there any reason this would change, if the only change was study->app?

We've flipped this back and forth several times in the past few months
https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/408
https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/411

It doesn't appear to fail in dev or staging. I think we should just leave this as is, and if it's an ongoing problem, we should file a Jira to track this down and figure it out once and for all.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityRecurringTest.java
alxdarksage: Yes this shouldn't have been changed, reverting.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityRecurringTest.java
```diff
@@ -143,19 +142,19 @@ public void retrievalActivitiesAcrossTimeAndTimeZones() throws Exception {
...
+        // filter it out from the persisted activities retrieved from the db.
         activities = filterList(usersApi.getScheduledActivities("-12:00", 2, null).execute().body(),
                 schedulePlan.getGuid());
+        
         assertEquals(4, activities.getItems().size());
         assertEquals(ytz1, activities.getItems().get(0).getScheduledOn().toString());
         assertEquals(ytz2, activities.getItems().get(1).getScheduledOn().toString());
         assertEquals(ytz3, activities.getItems().get(2).getScheduledOn().toString());
-        assertEquals(ytz4, activities.getItems().get(3).getScheduledOn().toString());
+        //assertEquals(ytz4, activities.getItems().get(3).getScheduledOn().toString());
```
DwayneJengSage, line 155: We should also restore this last assert and the comments as well.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ScheduledActivityRecurringTest.java
alxdarksage: I've reverted these changes, they were unintentional (I think they were there when I created the branch from some prior testing). Thanks for catching this.

DwayneJengSage (approved):

PR #426, opened 2020-05-08
Upload API can take non-encrypted, non-zipped uploads

DwayneJengSage:
https://github.com/Sage-Bionetworks/BridgeServer2/pull/202

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
@@ -452,17 +452,14 @@ public class UploadTest {
         request.setName(file.getName());
         request.setContentLength(file.length());
         request.setContentMd5(contentMd5);
+        request.setContentType("text/plain");
         request.setEncrypted(false);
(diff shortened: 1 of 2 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/UploadTest.java
```diff
@@ -436,6 +439,43 @@ public void miscTests() throws Exception {
...
+        UploadRequest request = new UploadRequest();
+        request.setName(file.getName());
+        request.setContentLength(file.length());
+        request.setContentMd5(contentMd5);
+        request.setEncrypted(false);
+        request.setZipped(false);
+
+        // We use application/zip for MIME type because that's what RestUtils expects. The MIME type doesn't actually
+        // matter for this test.
+        request.setContentType("application/zip");
```
alxdarksage, line 460: Should we change RestUtils to take the mime type?

alxdarksage (approved):
Approved but one question about making RestUtils more flexible, now that we can upload arbitrary file types.

PR #428, opened 2020-05-14
Test of CRC controller

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/CRCTest.java
```diff
@@ -0,0 +1,297 @@
...
+        assertEquals(user.getUserId(), retrieved.getIdentifier().get(0).getValue());
+        
+        StudyParticipant participant = adminUser.getClient(ParticipantsApi.class)
+                .getParticipantById(user.getUserId(), false).execute().body();
+        assertTrue(participant.getDataGroups().contains("tests_collected"));
+    }
+    
+    @Test
+    @Ignore
+    public void createObservation() throws IOException {
```
DwayneJengSage, line 252: Comment to document why this test is being ignored.

DwayneJengSage (approved):
Looks good. Need a comment on why that one test is being ignored.

src/test/java/org/sagebionetworks/bridge/sdk/integration/CRCTest.java
alxdarksage: Removed this @Ignore statement, this test is legit and passes.

PR #431, opened 2020-05-19
Verify testing support.

alxdarksage:
Verify that a test account, even when selected, does not call external partner and notifies the caller that it has done this.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/CRCTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/CRCTest.java
@@ -154,7 +154,7 @@ public class CRCTest {
     
     @Test
     public void updateParticipantWithDeclined() throws IOException {
-     // This is a normal call but I have not bothered to create a Java SDK method for
+        // This is a normal call but I have not bothered to create a Java SDK method for
         // it. Don't set 'selected' as a data group, it'll trigger a call to our external
         // partner.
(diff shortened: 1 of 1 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/CRCTest.java
```diff
@@ -123,17 +123,44 @@ public static void afterMethod() throws Exception {
         ExternalIdentifiersApi externalIdsApi = adminUser.getClient(ExternalIdentifiersApi.class);
         externalIdsApi.deleteExternalId(TEST_EXTERNAL_ID).execute();
     }
+
+    @Test
+    public void updateParticipantWithSelected() throws IOException {
+        // This is a normal call but I have not bothered to create a Java SDK method for
+        // it. Don't set 'selected' as a data group, it'll trigger a call to our external
+        // partner.
```
DwayneJengSage, line 131: Minor: This comment should be updated to reflect this particular test.

DwayneJengSage (approved):
Looks good. One comment about comments.

src/test/java/org/sagebionetworks/bridge/sdk/integration/CRCTest.java
alxdarksage: This is going to have to be revised given our design change yesterday. I will update BridgeServer2 and the integration tests with this change as part of this PR.

PR #456, opened 2020-08-10
Updates to reflect changes in the assessment resource API

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
```diff
@@ -223,28 +241,28 @@ public void test() throws Exception {
         sharedAssessmentsApi.importSharedAssessmentResource(id, ImmutableList.of(resourceGuid)).execute();
 
         resourcesPage = assessmentApi.getAssessmentResources(id, null, null, null, null, null, true).execute().body();
+        assertTrue(resourcesPage.getItems().stream().anyMatch(res -> res.getTitle().equals("This is a different shared title")));
```
DwayneJengSage, line 244: If we fail to clean up the test assessments, this title might match from a previous test run. Should we add a random token to the title to ensure that we have the resource that was created for this test run?

DwayneJengSage (approved):
Looks good. One minor comment.

src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
alxdarksage: Adding a random component to the title.

PR #461, opened 2020-08-26
Verify that deleted study does not show up in list of sponsored studies.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyTest.java
```diff
@@ -236,15 +236,30 @@ public void testSponsorship() throws Exception {
...
+            
+            adminStudiesApi.removeStudySponsor(STUDY_ID_1, ORG_ID_1).execute();
+
+            list = adminStudiesApi.getSponsors(STUDY_ID_1, null, null).execute().body();
+            assertFalse(list.getItems().stream().anyMatch((org) -> org.getIdentifier().equals(ORG_ID_1)));
+            
+        } finally {
+            // Need to recreate this for other tests, undeleted.
+            adminStudiesApi.deleteStudy(STUDY_ID_1, true).execute();
+        }
```
DwayneJengSage, line 263: Instead of modifying a study that's used by several tests, would it be better to create a study for this specific test?

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #484, opened 2020-11-13
Update to the latest Synapse Java client.

alxdarksage:
Wiped out my .m2 directory and then had to update this. All tests pass.


nategbrown9 (approved):
Looks good.

PR #488, opened 2020-11-30
Org administrator API tests

alxdarksage:
New APIs to create and manage administrative accounts, as a separate task from participants in studies.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountsTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountsTest.java
@@ -1,26 +1,39 @@
 package org.sagebionetworks.bridge.sdk.integration;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
+import static org.sagebionetworks.bridge.rest.model.AccountStatus.ENABLED;
 import static org.sagebionetworks.bridge.rest.model.Role.DEVELOPER;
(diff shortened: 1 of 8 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountsTest.java
```diff
@@ -0,0 +1,219 @@
...
+        AccountsApi accountsApi = emailUser.getClient(AccountsApi.class);
+        UserSessionInfo info = accountsApi.updateIdentifiersForSelf(identifierUpdate).execute().body();
+        assertEquals(PHONE.getNumber(), info.getPhone().getNumber());
+
+        Account retrieved = orgAdminApi.getAccount(emailUserId).execute().body();
+        assertEquals(PHONE.getNumber(), retrieved.getPhone().getNumber());
+
+        // But if you do it again, it should not work
+        info = accountsApi.updateIdentifiersForSelf(identifierUpdate).execute().body();
+        assertEquals(PHONE.getNumber(), info.getPhone().getNumber()); // unchanged
```
DwayneJengSage, line 217: The old version of the test used a different phone number. The new version of the test updates the phone number to itself, so we don't actually test anything interesting here.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/AccountsTest.java
alxdarksage: Changed to add a different phone number before trying again (also doesn't change the credential).

DwayneJengSage (approved):

PR #500, opened 2021-01-25
Cleanup after study coordinator merge.

alxdarksage:
As researchers become app-scoped, this one test needed to switch to a study coordinator to verify study-scoped access. Also changed BadRequestException to UnauthorizedException because...why was this a bad request exception?


nategbrown9 (approved):
Looks good.

PR #502, opened 2021-01-28
Enrollment deletion APIs and some additional tests

alxdarksage:
Verifies you can delete a custom event; verifies on server pathway that looked like it might be trouble (it isn't); made one test more resilient to other auto-custom events in the api app's config.


nategbrown9 (approved):
Looks good

PR #508, opened 2021-02-09
Added participant data integration tests.

kllysng:
Please let me know if I'm on the right track - I'm testing CRUD functionalities for the user and worker as well as the exception being thrown in response to a bad request.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
@@ -20,7 +20,6 @@ import java.util.HashSet;
 import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
@@ -130,7 +129,7 @@ public class ParticipantDataTest {
         assertEquals(3, results.getItems().size());
         ImmutableSet<String> expectedIdentifiers = ImmutableSet.of(identifier1, identifier2, identifier3);
         HashSet<String> actualIdentifiers = new HashSet<>(results.getItems());
-        assertFalse(actualIdentifiers.retainAll(expectedIdentifiers));
+        assertEquals(expectedIdentifiers, actualIdentifiers);
 
         // worker can get those participant data
         results = workersApi.getAllDataForAdminWorker(appId, userId, OFFSET_KEY, PAGE_SIZE).execute().body();
@@ -222,10 +221,10 @@ public class ParticipantDataTest {
         assertNull(pagedResults.getNextPageOffsetKey());
     }
 
-    private static void assertPages(ForwardCursorStringList pagedResults, String expectedIdentifier, int i) {
+    private static void assertPages(ForwardCursorStringList pagedResults, String expectedIdentifier, int start) {
         assertEquals(5, pagedResults.getItems().size());
-        for (; i < 5; i++) {
-            assertEquals(expectedIdentifier + i, pagedResults.getItems().get(i));
+        for (int i = start; i < 5 + start; i++) {
+            assertEquals(expectedIdentifier + i, pagedResults.getItems().get(i - start));
         }
     }
 
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
@@ -5,8 +5,11 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import org.sagebionetworks.bridge.rest.api.ForAdminsApi;
 import org.sagebionetworks.bridge.rest.api.ForConsentedUsersApi;
 import org.sagebionetworks.bridge.rest.api.ForWorkersApi;
+import org.sagebionetworks.bridge.rest.api.ParticipantDataApi;
+import org.sagebionetworks.bridge.rest.api.UsersApi;
 import org.sagebionetworks.bridge.rest.model.ParticipantData;
 import org.sagebionetworks.bridge.rest.model.StringList;
 import org.sagebionetworks.bridge.user.TestUserHelper;
@@ -15,9 +18,13 @@ import org.sagebionetworks.bridge.user.TestUserHelper.TestUser;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.sagebionetworks.bridge.rest.model.Role.DEVELOPER;
 import static org.sagebionetworks.bridge.rest.model.Role.WORKER;
 
 public class ParticipantDataTest {
@@ -32,6 +39,8 @@ public class ParticipantDataTest {
     private String identifier1;
     private String identifier2;
     private String identifier3;
+    private String offsetKey;
+    private Integer pageSize;
 
     @Before
     public void before() throws Exception {
@@ -40,6 +49,8 @@ public class ParticipantDataTest {
         identifier1 = Tests.randomIdentifier(ParticipantDataTest.class);
         identifier2 = Tests.randomIdentifier(ParticipantDataTest.class);
         identifier3 = Tests.randomIdentifier(ParticipantDataTest.class);
+        offsetKey = null;
+        pageSize = 10;
     }
 
     @After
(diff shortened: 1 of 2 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+        ForConsentedUsersApi usersApi = user.getClient(ForConsentedUsersApi.class);
+
+        StringList results = usersApi.getAllDataForSelf(offsetKey, pageSize).execute().body();
+        assertEquals(3, results.getItems().size());
+        HashSet<String> expectedIdentifiers = new HashSet<>(Arrays.asList(identifier1, identifier2, identifier3));
+        HashSet<String> actualIdentifiers = new HashSet<>(results.getItems());
+        assertFalse(actualIdentifiers.retainAll(expectedIdentifiers));
+
+        // worker can get those participant data
+        results = workersApi.getAllDataForAdminWorker(appId, userId, offsetKey, pageSize).execute().body();
```
DwayneJengSage, line 127: Also test getDataByIdentifierForAdminWorker()

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+        assertEquals(participantData1.getData(), actualData1.getData());
+        assertEquals(participantData2.getData(), actualData2.getData());
+        assertEquals(participantData3.getData(), actualData3.getData());
+
+        userApi.deleteDataByIdentifier(identifier1);
+        userApi.deleteDataByIdentifier(identifier2);
+        userApi.deleteDataByIdentifier(identifier3);
+
+        ForAdminsApi adminsApi = admin.getClient(ForAdminsApi.class);
+        adminsApi.deleteAllParticipantDataForAdmin(user.getAppId(), user.getUserId()).execute();
```
DwayneJengSage, line 98: Also test deleteDataForAdmin()

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+        ParticipantData participantData3 = createParticipantData("baz", "C");
+
+        userApi.saveDataForSelf(identifier1, participantData1).execute();
+        userApi.saveDataForSelf(identifier2, participantData2).execute();
+        userApi.saveDataForSelf(identifier3, participantData3).execute();
+
+        StringList results = userApi.getAllDataForSelf(offsetKey, pageSize).execute().body();
+
+        assertEquals(3, results.getItems().size());
+        HashSet<String> expectedIdentifiers = new HashSet<>(Arrays.asList(identifier1, identifier2, identifier3));
```
DwayneJengSage, line 82: Minor: You can use ImmutableSet.of(identifier1, identifier2, identifier3) to make this easier to write.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+        HashSet<String> actualIdentifiers = new HashSet<>(results.getItems());
+        assertFalse(actualIdentifiers.retainAll(expectedIdentifiers));
```
DwayneJengSage, line 84: You can use assertEquals() to check that two sets have the same values.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+        ParticipantData actualData1 = userApi.getDataByIdentifierForSelf(identifier1).execute().body();
+        ParticipantData actualData2 = userApi.getDataByIdentifierForSelf(identifier2).execute().body();
+        ParticipantData actualData3 = userApi.getDataByIdentifierForSelf(identifier3).execute().body();
+        assertEquals(participantData1.getData(), actualData1.getData());
+        assertEquals(participantData2.getData(), actualData2.getData());
+        assertEquals(participantData3.getData(), actualData3.getData());
+
+        userApi.deleteDataByIdentifier(identifier1);
+        userApi.deleteDataByIdentifier(identifier2);
+        userApi.deleteDataByIdentifier(identifier3);
```
DwayneJengSage, line 95: We should verify after this step that the data we deleted now no longer exists.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
DwayneJengSage, line 98: If the three previous calls to deleteDataByIdentifier() delete all the data, so this call doesn't actually do anything. Try using deleteDataByIdentifier() once and use this call to delete all the rest.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+        workersApi.saveDataForAdminWorker(appId, userId, identifier1, createParticipantData("foo", "A")).execute();
+        workersApi.saveDataForAdminWorker(appId, userId, identifier2, createParticipantData("bar", "B")).execute();
+        workersApi.saveDataForAdminWorker(appId, userId, identifier3, createParticipantData("baz", "C")).execute();
+
+        // user can get those participant data
+        ForConsentedUsersApi usersApi = user.getClient(ForConsentedUsersApi.class);
+
+        StringList results = usersApi.getAllDataForSelf(offsetKey, pageSize).execute().body();
+        assertEquals(3, results.getItems().size());
+        HashSet<String> expectedIdentifiers = new HashSet<>(Arrays.asList(identifier1, identifier2, identifier3));
```
DwayneJengSage, line 122: Similarly here.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+        HashSet<String> actualIdentifiers = new HashSet<>(results.getItems());
+        assertFalse(actualIdentifiers.retainAll(expectedIdentifiers));
+
+        // worker can get those participant data
+        results = workersApi.getAllDataForAdminWorker(appId, userId, offsetKey, pageSize).execute().body();
+        assertEquals(3, results.getItems().size());
+        actualIdentifiers = new HashSet<>(results.getItems());
+        assertFalse(actualIdentifiers.retainAll(expectedIdentifiers));
```
DwayneJengSage, line 130: Similarly here.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private static ParticipantData createParticipantData(String fieldValue1, String fieldValue2) {
+        ObjectNode node = JsonNodeFactory.instance.objectNode();
+        node.put("field1", fieldValue1);
+        node.put("field2", fieldValue2);
+        ParticipantData participantData = new ParticipantData();
+        participantData.setUserId(USER_ID);
+        participantData.setIdentifier(IDENTIFIER);
+        participantData.setData(String.valueOf(node));
```
DwayneJengSage, line 158: We'll want to change JavaSDK so that setData is an object instead of a String.

I'm not sure of the GSON library works with Jackson ObjectNodes. We might need to use a Map.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
DwayneJengSage, line 127: Let's also add pagination tests for getAllDataForSelf() and getAllDataForAdminWorker(). Something simple like creating 10 Participant Data and paginating through with page size 5 to make sure we get all 10 results.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
kellyyangsong: As I'm figuring this out, I'm realizing that because both of these return a ForwardCursorPagedResourceList<String>, that maybe I need to use something other than StringList because it appears to represent ResourceList<String> rather than a paged one in the swagger docs. I can't get the nextPageOffsetKey or page through results right now. 

Screen Shot 2021-02-11 at 4 18 22 PM

Should I make a new swagger doc for a ForwardCursorPagedResourceList<String>?

kellyyangsong: I just found paged_string.yml! This should do it.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,244 @@
...
+        workersApi.saveDataForAdminWorker(appId, userId, identifier3, createParticipantData("baz", "C")).execute();
+
+        // user can get those participant data
+        ForConsentedUsersApi usersApi = user.getClient(ForConsentedUsersApi.class);
+
+        ForwardCursorStringList results = usersApi.getAllDataForSelf(OFFSET_KEY, PAGE_SIZE).execute().body();
+        assertEquals(3, results.getItems().size());
+        ImmutableSet<String> expectedIdentifiers = ImmutableSet.of(identifier1, identifier2, identifier3);
+        HashSet<String> actualIdentifiers = new HashSet<>(results.getItems());
+        assertFalse(actualIdentifiers.retainAll(expectedIdentifiers));
```
DwayneJengSage, line 133: Missed one: This one can also be assertEquals()

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+        // check the remaining 5
+        pagedResults = workersApi.getAllDataForAdminWorker(user.getAppId(), user.getUserId(), nextKey, 5).execute().body();
+        assertPages(pagedResults, identifier1, 5);
+        assertNull(pagedResults.getNextPageOffsetKey());
+    }
+
+    private static void assertPages(ForwardCursorStringList pagedResults, String expectedIdentifier, int i) {
+        assertEquals(5, pagedResults.getItems().size());
+        for (; i < 5; i++) {
```
DwayneJengSage, line 227: This means that when i = 5, this loop does nothing. I think what you mean is the input should be start, and the for loop should be i < start + 5.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantDataTest.java
kellyyangsong: wow such a good catch

DwayneJengSage (approved):

alxdarksage (approved):

PR #509, opened 2021-02-20
Refactor for security changes.

alxdarksage:
Some exception types changed, and roles switched to maintain study-scoping/authorization tests using a study coordinator (rather than a researcher).

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyFilteringTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyFilteringTest.java
@@ -189,7 +189,7 @@ public class StudyFilteringTest {
         assertListContainsAccount(list.getItems(), STUDY_ID_1, user1.getId());
         assertListContainsAccount(list.getItems(), STUDY_ID_1, user1and2.getId());
         assertListContainsAccount(list.getItems(), STUDY_ID_2, user1and2.getId());
-        assertListContainsAccount(list.getItems(), STUDY_ID_1, user2.getId());
+        assertListContainsAccount(list.getItems(), STUDY_ID_2, user2.getId());
         
         // Researcher 2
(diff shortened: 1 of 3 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyFilteringTest.java
```diff
@@ -120,53 +127,86 @@ public static void after() throws Exception {
...
+        ClientManager manager = new ClientManager.Builder().withSignIn(researcher1).build();
+        ForResearchersApi researchersApiFor1 = manager.getClient(ForResearchersApi.class);
+        
+        // Researcher 1 sees all users
+        AccountSummaryList list = researchersApiFor1.searchAccountSummaries(
+                new AccountSummarySearch()).execute().body();
+        assertListContainsAccount(list.getItems(), STUDY_ID_1, user1.getId());
+        assertListContainsAccount(list.getItems(), STUDY_ID_1, user1and2.getId());
+        assertListContainsAccount(list.getItems(), STUDY_ID_2, user1and2.getId());
+        assertListContainsAccount(list.getItems(), STUDY_ID_1, user2.getId());
```
DwayneJengSage, line 192: Should this say STUDY_ID_2?

DwayneJengSage (commented):
Looks good. One question about a test.

src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyFilteringTest.java
alxdarksage: Yes, changed. However the caller is a researcher, so it's not surprising STUDY_ID_1 also works, but it should be study 2 for completeness.

DwayneJengSage (approved):

PR #533, opened 2021-05-21
User note

jnelsonjava:
- Updated AccountsTest and ParticipantsTest to verify note field


jnelsonjava:
Should check for AccountSummary.

alxdarksage (approved):

PR #542, opened 2021-06-16
Permission changes for study coordinators

alxdarksage:
One of the bigger changes is that participants can get any study under their ID irrespective of their current enrollments...an edge case that will probably never come up.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
@@ -6,13 +6,11 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 import static org.sagebionetworks.bridge.rest.model.ResourceCategory.DATA_REPOSITORY;
 import static org.sagebionetworks.bridge.rest.model.ResourceCategory.LICENSE;
 import static org.sagebionetworks.bridge.rest.model.ResourceCategory.WEBSITE;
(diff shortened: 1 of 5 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
```diff
@@ -6,13 +6,11 @@
 import static org.junit.Assert.assertNotEquals;
```
alxdarksage, line 6: Why were these tests entirely removed, and not updated for clarity like the next test?

src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentTest.java
```diff
@@ -92,7 +92,7 @@ public void before() throws Exception {
         orgsApi.addMember(ORG_ID_2, otherDeveloper.getUserId()).execute();
```
alxdarksage, line 92: Org membership shouldn't matter if you're a developer, but it should if you're a study designer. So what's happening here?

src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentTest.java
```diff
@@ -256,13 +238,6 @@ public void test() throws Exception {
         // deleteAssessment physical=false works
         assessmentApi.deleteAssessment(secondRevision.getGuid(), false).execute();
         
-        // deleteAssessment physical=false fails for developer who doesn't own the assessment
-        try {
-            badDevApi.deleteAssessment(secondRevision.getGuid(), false).execute();
-            fail("Should have thrown an exception");
-        } catch(EntityNotFoundException e) {
-        }
```
alxdarksage, line 264: So verify that the ownerId was changed to the caller's organization

PR #544, opened 2021-06-17
Fix tests for permissions changes

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
```diff
@@ -181,97 +192,94 @@ public void test() throws Exception {
         assertFalse(rp.isIncludeDeleted());
 
         // Other developers can see local resources, just as they can see the assessment
-        resourcesPage = badDevApi.getAssessmentResources(
+        resourcesPage = assessmentApiOrg2.getAssessmentResources(
```
alxdarksage, line 195: where is assessmentApiOrg2.getAssessment() being tested for asme owned by org1?

src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
```diff
@@ -293,4 +301,150 @@ public void test() throws Exception {
...
+        admin.getClient(SharedAssessmentsApi.class).deleteSharedAssessmentResource(
+                id, sharedResource.getGuid(), false).execute();
+        sharedCount = sharedAssessmentsApiOrg1.getSharedAssessmentResources(
+                id, null, null, null, null, null, false).execute().body().getTotal();
+        assertEquals(0, sharedCount);
+        sharedCount = sharedAssessmentsApiOrg1.getSharedAssessmentResources(
+                id, null, null, null, null, null, true).execute().body().getTotal();
+        assertEquals(1, sharedCount);
+        sharedResource = sharedAssessmentsApiOrg1.getSharedAssessmentResource(id, sharedResource.getGuid()).execute().body();
+        assertNotNull(sharedResource);
```
alxdarksage, line 448: You can still probably delete all of this, all you're testing is the role security change.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -235,12 +235,12 @@ public void workerCanCrudParticipantReportByDateTime() throws Exception {
...
-        workerReportsApi.addParticipantReportRecord(reportId, makeReportDataForWorker(healthCode, DATETIME2, "bar",
-                "B")).execute();
-        workerReportsApi.addParticipantReportRecord(reportId, makeReportDataForWorker(healthCode, DATETIME3, "baz",
-                "C")).execute();
+        workerReportsApi.addParticipantReportRecordForWorker(
+                reportId, makeReportDataForWorker(healthCode, DATETIME1, "foo", "A")).execute();
+        workerReportsApi.addParticipantReportRecordForWorker(
+                reportId, makeReportDataForWorker(healthCode, DATETIME2, "bar", "B")).execute();
+        workerReportsApi.addParticipantReportRecordForWorker(
+                reportId, makeReportDataForWorker(healthCode, DATETIME3, "baz", "C")).execute();
```
alxdarksage, line 243: Double-check, is this like other methods in the WorkerApi class?

src/test/java/org/sagebionetworks/bridge/sdk/integration/AssessmentResourceTest.java
alxdarksage: On second thought, let's leave it to verify the study designer can go through all the steps to publish/import/search for/create/delete assessments like a developer.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: Yes, we do have to do this for some calls that otherwise would be named the same, and this is one of them.

alxdarksage: This occurred because I added a v3 method to add participant reports, but it is in fact redundant with the v4 endpoint. Removed it, and reverted the name of this method to remove "ForWorker".

nategbrown9 (approved):
Looks good.

PR #547, opened 2021-06-24
Verifying previous S3 file is deleted on create request

jnelsonjava:


src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
```diff
@@ -130,6 +130,14 @@ public void crudParticipantFile() throws Exception {
 
         // Updating an existing file
         ParticipantFile updateKeys = userApi.createParticipantFile("file_id", file).execute().body();
+
+        try {
+            userApi.getParticipantFile("file_id").execute().body();
+            fail("Previous S3 file should have been deleted by new Create request.");
+        } catch (Exception exception) {
+            assertEquals("Not Found", exception.getMessage());
```
DwayneJengSage, line 138: What kind of exception gets thrown here? Ideally it should be an EntityNotFoundException, or at least a BridgeSDKException with status code 404.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
jnelsonjava: It is a BridgeSDKException on this one. The GET route throws an EntityNotFoundException if the ParticipantFile record is missing, but it seems to throw a BridgeSDKException on the redirect. Would it be better to assert the type of exception?

DwayneJengSage: Hmm, I think the cleanest thing to do is in BridgeServer, catch the Amazon exception, check if it's a Not Found, and then re-throw it as a Bridge EntityNotFoundException. This would make our JavaSDK throw an EntityNotFoundException, which would make it cleaner.

alxdarksage: I think the issue is that the method returns a 302 and you only get the 404 when the HTTP client redirects to S3. So this is a translation of an S3 return status. I would at least catch BridgeSDKException to verify we're catching this, and it should wrap the underlying exception, whatever it is (probably an OkHttp client exception)? We can verify that.

alxdarksage (changes requested):
See comment

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
DwayneJengSage: Excellent point. I guess we should figure out the exact mechanism of the exception being thrown here. I assumed that our call to S3.generatePresignedUrl() was throwing within BridgeServer. But if S3 is just generating a link that points to nothing, then it would be the redirect that's throwing, and we wouldn't be able to intercept that and replace it with a Bridge EntityNotFoundException. In this case, it would be some kind of HTTP Exception with status 404. Might be worthwhile stepping through this in a debugger and seeing what's happening.

Other open questions: Does the HTTP tool we use in Integ Tests automatically follow redirects? I know that Postman has a toggle for follow redirects.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
```diff
@@ -130,6 +131,17 @@ public void crudParticipantFile() throws Exception {
...
+
+        try {
+            userApi.getParticipantFile("file_id").execute().body();
+            fail("Previous S3 file should have been deleted by new Create request.");
+        } catch (BridgeSDKException exception) {
+            assertEquals("Not Found", exception.getMessage());
+            assertEquals(404, exception.getStatusCode());
+        } catch (Exception e) {
+            fail("Unexpected Exception type.");
+        }
```
alxdarksage, line 143: This second catch is unnecessary. If the exception is not BridgeSDKException, it will be thrown out of the method, causing the test to fail.

alxdarksage (changes requested):
One minor change (I think you are still looking at what the underlying exception is that we're receiving back when the file doesn't exist on S3?)

alxdarksage (approved):

DwayneJengSage (approved):

PR #567, opened 2021-08-17
Define events on a study, not the app

alxdarksage:
BRIDGE-3044. Automatic custom events will be replaced with some kind of support for study bursts, that will probably generate these events dynamically without persisting them.


src/test/java/org/sagebionetworks/bridge/sdk/integration/ActivityEventTest.java
```diff
@@ -8,9 +8,9 @@
 import static org.junit.Assert.assertNull;
```
alxdarksage, line 8: Could this be StudyActivityEventTest and then we have an ActivityEventTest that tests the global (app-scoped) event system?

alxdarksage: Separated these out and restored the code manipulating the app scope.

jnelsonjava (approved):

PR #568, opened 2021-08-18
Updates based on multiple events in session, with changes to timeline.

alxdarksage:
BRIDGE-3059.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.22.24</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.22.23</version>
+            <version>0.22.24</version>
             <scope>test</scope>
         </dependency>
         <dependency>
```
alxdarksage, line 65: Add a test where a schedule with multiple start event IDs, generates the correct entries in the timeline.

jnelsonjava (approved):

PR #571, opened 2021-09-10
Test changes for developer access to test accounts

alxdarksage:
BRIDGE-3053

Developers must now operate on accounts that are marked as test accounts. For some calls, this wasn't previously necessary.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthorizationTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthorizationTest.java
@@ -1,0 +1,232 @@
+package org.sagebionetworks.bridge.sdk.integration;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.sagebionetworks.bridge.rest.model.Role.DEVELOPER;
+import static org.sagebionetworks.bridge.rest.model.Role.RESEARCHER;
+import static org.sagebionetworks.bridge.rest.model.Role.STUDY_COORDINATOR;
+import static org.sagebionetworks.bridge.rest.model.Role.STUDY_DESIGNER;
+import static org.sagebionetworks.bridge.sdk.integration.Tests.STUDY_ID_1;
+import static org.sagebionetworks.bridge.util.IntegTestUtils.SAGE_ID;
+
+import java.util.concurrent.Callable;
+
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.sagebionetworks.bridge.rest.api.ForOrgAdminsApi;
+import org.sagebionetworks.bridge.rest.api.ParticipantsApi;
+import org.sagebionetworks.bridge.rest.api.StudiesApi;
+import org.sagebionetworks.bridge.rest.api.StudyParticipantsApi;
+import org.sagebionetworks.bridge.rest.model.AccountSummaryList;
+import org.sagebionetworks.bridge.rest.model.AccountSummarySearch;
+import org.sagebionetworks.bridge.rest.model.Enrollment;
+import org.sagebionetworks.bridge.rest.model.StudyParticipant;
+import org.sagebionetworks.bridge.user.TestUserHelper;
+import org.sagebionetworks.bridge.user.TestUserHelper.TestUser;
+
+import retrofit2.Call;
+import retrofit2.Response;
+
+public class AuthorizationTest {
+
+    private static TestUser developer;
+    private static TestUser researcher;
+    private static TestUser studyDesigner;
+    private static TestUser studyCoordinator;
+    
+    private TestUser prodUser;
+    private TestUser testUser;
+    
+    private String prodUserId;
+    private String prodUserEmail;
+    private String testUserId;
+    private String testUserEmail;
+
+    @BeforeClass
+    public static void beforeTests() throws Exception {
+        developer = TestUserHelper.createAndSignInUser(AuthorizationTest.class, false, DEVELOPER);
+        researcher = TestUserHelper.createAndSignInUser(AuthorizationTest.class, false, RESEARCHER);
+        studyDesigner = TestUserHelper.createAndSignInUser(AuthorizationTest.class, false, STUDY_DESIGNER);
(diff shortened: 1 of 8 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantsTest.java
```diff
@@ -598,7 +598,9 @@ public void getActivityHistory() throws Exception {
     
     @Test
     public void getActivityHistoryV4() throws Exception {
-        TestUser user = TestUserHelper.createAndSignInUser(ParticipantsTest.class, true, Role.DEVELOPER);
+        TestUser user = new TestUserHelper.Builder(ParticipantsTest.class)
+                .withRoles(DEVELOPER).withConsentUser(true).isTestUser().createAndSignInUser();
+
         ForConsentedUsersApi usersApi = user.getClient(ForConsentedUsersApi.class);
         
         SchedulesV1Api schedulePlanApi = user.getClient(SchedulesV1Api.class);
```
alxdarksage, line 606: Let's add an integration test of the permissions structure itself, not just as a side effect of set up for other tests. 

Also here, we should have really split out the developer and the test account being operated on. But check why developer couldn't operate on self.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -137,7 +137,9 @@ public static void deleteDeveloper() throws Exception {
 
     @Test
     public void developerCanCrudParticipantReport() throws Exception {
-        user = TestUserHelper.createAndSignInUser(ReportTest.class, true);
+        user = new TestUserHelper.Builder(ReportTest.class)
+                .withConsentUser(true)
+                .isTestUser().createAndSignInUser();
```
alxdarksage, line 142: Rename to "asTestUser"

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -186,7 +188,9 @@ public void developerCanCrudParticipantReport() throws Exception {
 
     @Test
     public void workerCanCrudParticipantReportByDate() throws Exception {
-        user = TestUserHelper.createAndSignInUser(ReportTest.class, true);
+        user = new TestUserHelper.Builder(ReportTest.class)
+                .withConsentUser(true)
+                .isTestUser().createAndSignInUser();
```
alxdarksage, line 193: If this is a worker, then the account does not need to be a test account.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -226,7 +230,9 @@ public void workerCanCrudParticipantReportByDate() throws Exception {
 
     @Test
     public void workerCanCrudParticipantReportByDateTime() throws Exception {
-        user = TestUserHelper.createAndSignInUser(ReportTest.class, true);
+        user = new TestUserHelper.Builder(ReportTest.class)
+                .withConsentUser(true)
+                .isTestUser().createAndSignInUser();
```
alxdarksage, line 235: Again, worker.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
```diff
@@ -399,7 +405,10 @@ public void correctExceptionsOnBadRequest() throws Exception {
     
     @Test
     public void userCanCRUDSelfReports() throws Exception {
-        user = TestUserHelper.createAndSignInUser(ReportTest.class, true);
+        user = new TestUserHelper.Builder(ReportTest.class)
+                .withConsentUser(true)
+                .isTestUser().createAndSignInUser();
```
alxdarksage, line 410: Same. This test needs to succeed for a production account too.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: It was marked as a test account because a developer account was used to delete the associated reports. For this reason, I've removed the test_user flag here, and used an admin to do test cleanup.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: Changing cleanup removes the need to make a test account.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: Removed after fixing cleanup.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ReportTest.java
alxdarksage: renamed to "withTestDataGroup()"

alxdarksage:
In addition to verifying these changes were for cleanup, and switching to the user of an admin account so we don't have to add "test_user" to accounts at random, I also added a test to more systematically verify security for getting an individual account and for searching.

src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthorizationTest.java
```diff
@@ -0,0 +1,278 @@
...
+        StudyParticipant testParticipant = resPartApi.getParticipantById(testUserId, false).execute().body();
+        StudyParticipant prodParticipant = resPartApi.getParticipantById(prodUserId, false).execute().body();
+     
+        // Developers can only update test accounts.
+        shouldFail(() -> devPartApi.updateParticipant(prodUserId, prodParticipant));
+        devPartApi.updateParticipant(testUserId, testParticipant).execute();
+
+        // Researchers can update both kinds of accounts
+        resPartApi.updateParticipant(prodUserId, prodParticipant);
+        resPartApi.updateParticipant(testUserId, testParticipant);
```
DwayneJengSage, line 220: These are missing .execute(). Recommend double-checking to make sure all the calls have their .execute().

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/AuthorizationTest.java
alxdarksage: Added and checked other methods for .execute().

DwayneJengSage (approved):

PR #579, opened 2021-10-05
Add enrollments to sessions

alxdarksage:
BRIDGE-3097

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
@@ -575,7 +575,7 @@ public class ConsentTest {
             // verify the participant is now not enrolled in any study
             StudyParticipant participant = participantsApi.getParticipantById(user.getUserId(), false).execute().body();
             assertNull(participant.getEnrollments().get(STUDY_ID_1).isWithdrawnBySelf());
-            assertNotNull(researchUser.getUserId(), participant.getEnrollments().get(STUDY_ID_1).getWithdrawnOn());
+            assertNotNull(participant.getEnrollments().get(STUDY_ID_1).getWithdrawnOn());
         });
     }
     
@@ -592,7 +592,7 @@ public class ConsentTest {
             StudyParticipant participant = participantsApi.getParticipantById(user.getUserId(), false).execute().body();
             assertNull(participant.getEmail());
             assertNull(participant.getEnrollments().get(STUDY_ID_1).isWithdrawnBySelf());
-            assertNotNull(researchUser.getUserId(), participant.getEnrollments().get(STUDY_ID_1).getWithdrawnOn());
+            assertNotNull(participant.getEnrollments().get(STUDY_ID_1).getWithdrawnOn());
         });
     }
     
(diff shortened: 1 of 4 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
```diff
@@ -555,6 +558,41 @@ public void consentAndWithdrawFromAppUpdatesDataGroupsAndStudies() throws Except
...
+            Enrollment en = researchersApi.withdrawParticipant(STUDY_ID_1, user.getUserId(), "Withdrawn").execute().body();
+            
+            assertEquals(researchUser.getUserId(), en.getWithdrawnBy());
+            
+            ParticipantsApi participantsApi = researchUser.getClient(ParticipantsApi.class);
+
+            // verify the participant is now not enrolled in any study
+            StudyParticipant participant = participantsApi.getParticipantById(user.getUserId(), false).execute().body();
+            assertNull(participant.getEnrollments().get(STUDY_ID_1).isWithdrawnBySelf());
+            assertNotNull(researchUser.getUserId(), participant.getEnrollments().get(STUDY_ID_1).getWithdrawnOn());
```
alxdarksage, line 578: These should be two separate tests

src/test/java/org/sagebionetworks/bridge/sdk/integration/ConsentTest.java
```diff
@@ -555,6 +558,41 @@ public void consentAndWithdrawFromAppUpdatesDataGroupsAndStudies() throws Except
...
+            researchersApi.withdrawParticipantFromApp(user.getUserId(), 
+                    new Withdrawal().reason("Withdrawn")).execute().body();
+            
+            ParticipantsApi participantsApi = researchUser.getClient(ParticipantsApi.class);
+
+            // verify the participant is now not enrolled in any study
+            StudyParticipant participant = participantsApi.getParticipantById(user.getUserId(), false).execute().body();
+            assertNull(participant.getEmail());
+            assertNull(participant.getEnrollments().get(STUDY_ID_1).isWithdrawnBySelf());
+            assertNotNull(researchUser.getUserId(), participant.getEnrollments().get(STUDY_ID_1).getWithdrawnOn());
```
alxdarksage, line 595: These should be two separate tests

jnelsonjava (approved):
Approved with just the one comment to check out.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantsTest.java
```diff
@@ -201,6 +206,11 @@ public void retrieveParticipant() throws Exception {
             assertEquals(user.getSession().getId(), participant.getId());
             assertFalse(participant.getConsentHistories().isEmpty());
             
+            assertEquals(user.getSession().getEnrollments().get(STUDY_ID_1).getEnrolledOn(), 
+                    participant.getEnrollments().get(STUDY_ID_1).getEnrolledOn());
+            assertEquals(user.getSession().getEnrollments().size(), participant.getEnrollments().size());
+            assertEquals((String)null, (String)null);
```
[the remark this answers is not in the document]
alxdarksage: Sorry that was a test to see if asserEquals would return true if both values were null, which apparently it does. So it alerted me to the fact that a test like assertEquals(foo.getValue(), bar.getValue()) might succeed even though the value isn't there. Removed.

PR #584, opened 2021-10-12
Bridge to talk to Synapse Dev

DwayneJengSage:
See https://github.com/Sage-Bionetworks/BridgeServer2/pull/449

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
@@ -80,6 +80,8 @@ public class Tests {
     private static final String DEFAULT_CONFIG_FILE = CONFIG_FILE;
     private static final String USER_CONFIG_FILE = System.getProperty("user.home") + "/" + CONFIG_FILE;
 
+    private static Config config;
+
     public static ClientInfo getClientInfoWithVersion(String osName, int version) {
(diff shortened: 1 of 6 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
```diff
@@ -329,7 +341,35 @@ public static void assertDatesWithTimeZoneEqual(DateTime expected, DateTime actu
...
+    public static Config loadTestConfig() throws IOException {
+        Config config;
+        Path localConfigPath = Paths.get(USER_CONFIG_FILE);
+        if (Files.exists(localConfigPath)) {
+            config = new PropertiesConfig(DEFAULT_CONFIG_FILE, localConfigPath);
+        } else {
+            config = new PropertiesConfig(DEFAULT_CONFIG_FILE);
+        }
+        return config;
+    }
```
alxdarksage, line 354: If we can load the config statically, then tests like AppTest could just reference it rather than loading the file again. It probably doesn't matter much.

alxdarksage (approved):

src/test/java/org/sagebionetworks/bridge/sdk/integration/Tests.java
DwayneJengSage: I did the next best thing: I cached the config, so if you call loadConfig() again, it re-uses it.

PR #607, opened 2021-11-17
Verifying study burst information in the timeline.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyBurstTest.java
```diff
@@ -11,6 +11,7 @@
 import static org.sagebionetworks.bridge.util.IntegTestUtils.SAGE_ID;
 import static org.sagebionetworks.bridge.util.IntegTestUtils.TEST_APP_ID;
 
+import java.math.BigDecimal;
```
alxdarksage, line 14: You never want to see this as the output from Swagger

alxdarksage: It was an unused import, removed.

PR #610, opened 2021-11-24
Add test user flag for design studies

alxdarksage:
Also taking the opportunity to eliminate one kind fo copy-and-paste error by replacing Tests.randomIdentifier(MyTest.class) with Tests.randomIdentifier(getClass()).

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/PersistentActivityTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/PersistentActivityTest.java
@@ -64,7 +64,7 @@ public class PersistentActivityTest {
     @SuppressWarnings("deprecation")
     @Test
     public void persistentActivityCanBeFinishedMultipleTimesInDay() throws Exception {
-        String activityLabel1 = Tests.randomIdentifier(this.getClass());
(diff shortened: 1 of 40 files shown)


nategbrown9 (approved):
Looks good.

src/test/java/org/sagebionetworks/bridge/sdk/integration/PersistentActivityTest.java
```diff
@@ -64,7 +64,7 @@ public void after() throws Exception {
     @SuppressWarnings("deprecation")
     @Test
     public void persistentActivityCanBeFinishedMultipleTimesInDay() throws Exception {
-        String activityLabel1 = Tests.randomIdentifier(this.getClass());
+        String activityLabel1 = Tests.randomIdentifier(getClass());
```
jnelsonjava, line 67: I think making this change will convert the class found from PersistentActivityTest to String. I don't think it would affect the outcome of the test, but just wanted to check that's what you intended.

alxdarksage: this.getClass() and getClass() are equivalent. Both return Class.

jnelsonjava (commented):
Looks good!

PR #614, opened 2021-12-06
BRIDGE-3152 (integration tests in production)

alxdarksage:
- Changed all tests to use preferred methods that replace the two removed superadmin APIs;
- Changed smoke tests so they only need an admin to run (removed UTF-8 test because it's definitely not necessary in the smoke tests, and added a couple of others that seemed important, but I'm flexible about which tests we include in this set).

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/InitListener.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/InitListener.java
@@ -151,19 +151,18 @@ public class InitListener extends RunListener {
             admin.getClient(ForOrgAdminsApi.class).addMember(SAGE_ID, admin.getUserId()).execute();
         }
         
-        admin.getClient(ForSuperadminsApi.class).adminChangeApp(new SignIn().appId(SHARED_APP_ID)).execute();
-        
-        try {
-            orgsApi.getOrganization(SAGE_ID).execute();
-        } catch(EntityNotFoundException e) {
-            Organization org = new Organization().identifier(SAGE_ID).name(SAGE_NAME)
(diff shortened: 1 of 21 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/InitListener.java
```diff
@@ -151,19 +148,18 @@ public void testRunStarted(Description description) throws Exception {
...
+//        try {
+//            orgsApi.getOrganization(SAGE_ID).execute();
+//        } catch(EntityNotFoundException e) {
+//            Organization org = new Organization().identifier(SAGE_ID).name(SAGE_NAME)
+//                    .description("Sage sponsors study1 and study2");
+//            orgsApi.createOrganization(org).execute();
+//            LOG.info("  Creating organization “{}” in shared study", SAGE_ID);
+//        } finally {
+//            admin.getClient(ForSuperadminsApi.class).adminChangeApp(new SignIn().appId(TEST_APP_ID)).execute();
+//        }
```
alxdarksage, line 162: Remove this comment. Happening now in InitListener.

DwayneJengSage: This _is_ InitListener. Was this supposed to be deleted?

DwayneJengSage (approved):

src/test/java/org/sagebionetworks/bridge/sdk/integration/InitListener.java
alxdarksage: No, this was commented out while testing. Restored this, and switched over to use the changeApp API call. Tests still all pass.

PR #616, opened 2021-12-11
Create the event stream adherence report

alxdarksage:


pom.xml
```diff
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.23.11</version>
+            <version>0.23.12</version>
```
jnelsonjava, line 62: I think this should be 0.23.14

jnelsonjava (approved):
Looks good but check the SDK version before merging.

pom.xml
alxdarksage: Updated

PR #621, opened 2021-12-14
Adjusting tests based on using an admin who signs in via Synapse

alxdarksage:
One test (OAuthForWorkerTest) I had to entirely disable for now, as we'll need a valid Synapse ID for the worker being created for the test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 # BridgeIntegrationTests
 
-Integration tests for the BridgeServer2 server. Integration tests require a bootstrap account on the server to execute. Outside of production, the full suite of tests runs using  `SUPERADMIN` account. On production, for security reasons, a subset of these tests execute using an `ADMIN` account. (The `ADMIN` account can’t access all apps, or create worker accounts that can access all apps.)
-
-The production tests are marked with the `@Category(IntegrationSmokeTest.class)` class annotation.
+Integration tests for the BridgeServer2 server. Integration tests require bootstrap accounts on the server to execute. All tests run outside of production using a `SUPERADMIN` account. In production, for security reasons, a subset of these tests execute using an `ADMIN` account (marked with the `@Category(IntegrationSmokeTest.class)` class annotation). The `ADMIN` account can’t access all apps, or create worker accounts that can access all apps.
 
 ## Setting up your environment to run integration tests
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -18,16 +16,13 @@ The integration tests use the Bridge Java REST SDK, and both the SDK and the tes
...
-    admin.email = <email address>
-    admin.password = <password>
-
     synapse.test.user = <user name>
     synapse.test.user.id = <numerical account ID>
     synapse.test.user.password = <password>
     synapse.test.user.api.key = <api key>
 
-**admin.email** and **admin.password:** Create an account with the `SUPERADMIN` role outside of production and the `ADMIN` role in production. This account should be created in the 'api' and 'shared' apps, with the same email and password, and the same synapse user ID of 0000 (four zeroes) in both apps (this allows an `ADMIN` account to to switch between these two studies). These are test apps, so accidents in these apps would not impact other production apps.
+Create two bootstrap accounts in each environment, one in the 'api' app and one in the 'shared' app. Each account should have the `synapseUserId` ID set in the `synapse.test.user.id` property. In production these accounts should have the `ADMIN` role and in other environments, these accounts should have the `SUPERADMIN` role.
```
DwayneJengSage, line 24: Do we need to update the DefaultAppBootstrapper to do this?

src/test/java/org/sagebionetworks/bridge/sdk/integration/OAuthTest.java
```diff
@@ -84,15 +78,16 @@ public void requestOAuthAccessTokenExists() throws Exception {
     @Test
     public void signInWithSynapseAccount() throws Exception {
         String oauthClientId = CONFIG.get("synapse.oauth.client.id");
+        String synapseEndpoint = CONFIG.get("synapse.endpoint");
+        if (CONFIG.getEnvironment() == Environment.PRODUCTION) {
+            oauthClientId = CONFIG.get("prod.synapse.oauth.client.id");
+            synapseEndpoint = CONFIG.get("prod.synapse.endpoint");
+        }
```
DwayneJengSage, line 85: Does Config not handle this for us? It did before when I originally wrote this.

DwayneJengSage (approved):

src/test/java/org/sagebionetworks/bridge/sdk/integration/OAuthTest.java
alxdarksage: I switched to the use of the Config implementation in the integration test utilities. I have a JIRA to remove that entirely and have everything use the bridge-base Config, in which case, this can be removed. But that bridge-base implementation will need to be able to load multiple property files, as the SDK implementation.

README.md
alxdarksage: We could, yes. Maybe we could keep the email/password but create that admin account with the synapseUserId it finds in the property files, that would allow the integration tests to pass. I'll add this to BRIDGE-3169.

alxdarksage: Added to BRIDGE-3169 for follow-up.

src/test/java/org/sagebionetworks/bridge/sdk/integration/OAuthTest.java
alxdarksage: Added to BRIDGE-3169 for follow-up.

PR #647, opened 2022-02-22
Etag support (BRIDGE-3224)

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/EtagsTest.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    @After
+    public void after() throws Exception {
+        TestUser admin = TestUserHelper.getSignedInAdmin();
+        SchedulesV2Api adminSchedulesApi = admin.getClient(SchedulesV2Api.class);
+        
+        if (schedule != null) {
+            adminSchedulesApi.deleteSchedule(schedule.getGuid()).execute();    
+        }
```
DwayneJengSage, line 56: Minor: If saveScheduleForStudy() fails, this will probably throw an error on delete too. Perhaps check that schedule.getGuid() is also non-null as well?

src/test/java/org/sagebionetworks/bridge/sdk/integration/EtagsTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertEquals(statusCode, response.getStatusLine().getStatusCode());        
+    }
+
+    private void assertUserStatus(String url, String etag, int statusCode) throws IOException { 
+        HttpResponse response = Request.Get(hostUrl + url)
+            .setHeader("Bridge-Session", user.getSession().getSessionToken())
+            .setHeader(HttpHeaders.IF_NONE_MATCH, etag)
+            .execute().returnResponse();
+        assertEquals(statusCode, response.getStatusLine().getStatusCode());        
+    }
```
DwayneJengSage, line 124: These two methods differ only by studyDesigner vs user. Would it be cleaner to make this a single method that takes in the TestUser as an input?

DwayneJengSage (approved):
Looks good. Very cool. A few minor comments.

src/test/java/org/sagebionetworks/bridge/sdk/integration/EtagsTest.java
alxdarksage: Changed to check for the GUID as well.

DwayneJengSage (approved):

PR #649, opened 2022-02-24
Revising participant schedule (BRIDGE-3221)

alxdarksage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/Schedule2Test.java
```diff
@@ -508,26 +505,17 @@ public void getParticipantScheduleForStudyParticipant() throws Exception {
...
         assertEquals(7, schedule.getSchedule().size());
         
-        // Let's try and change the time zone and see what happens
-        DateTime now = DateTime.now();
-        schedule = userApi.getParticipantScheduleForSelf(STUDY_ID_1, EUROPE_PARIS_TZ).execute().body();
-        assertEquals(EUROPE_PARIS_TZ, schedule.getClientTimeZone());
-        assertEquals(DateTimeZone.forID(EUROPE_PARIS_TZ).getOffset(now), schedule.getCreatedOn().getZone().getOffset(now));
-        
-        StudyParticipant participant = userApi.getUsersParticipantRecord(false).execute().body();
-        assertEquals(EUROPE_PARIS_TZ, participant.getClientTimeZone());
```
alxdarksage, line 523: Add a test that 304 works here, rather than 'if-modified-since' header that was being used before.

alxdarksage: Updated this, and discovered caching was always failing because the timestamp was being updated whether the the timeline_retrieved event was actually being updated or not. Changed the server code so that two requests to this endpoint leads to a 304 (the first time an event is recorded as a side effect, so the tag that is returned changes on the next request).

DwayneJengSage (approved):

PR #687, opened 2022-07-05
Add integration test for participant file download rate limiting

imranmaj:
[https://sagebionetworks.jira.com/browse/BRIDGE-2941](https://sagebionetworks.jira.com/browse/BRIDGE-2941)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>org.sagebionetworks</groupId>
     <artifactId>BridgeIntegrationTests</artifactId>
-    <version>0.9.0</version>
+    <version>0.9.1</version>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>org.sagebionetworks</groupId>
     <artifactId>BridgeIntegrationTests</artifactId>
-    <version>0.9.1</version>
+    <version>0.9.0</version>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>rest-client</artifactId>
-            <version>0.25.7</version>
+            <version>0.25.8</version>
         </dependency>
         <dependency>
             <groupId>org.sagebionetworks</groupId>
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
@@ -1,6 +1,7 @@
 package org.sagebionetworks.bridge.sdk.integration;
 
 import okhttp3.ResponseBody;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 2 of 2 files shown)


jnelsonjava (approved):
Approved. But one minor addition. Since the SDK version is bumping, also bump the bridge rest-client in the integration test's pom.xml to match.

pom.xml
```diff
@@ -5,7 +5,7 @@
 
     <groupId>org.sagebionetworks</groupId>
     <artifactId>BridgeIntegrationTests</artifactId>
-    <version>0.9.0</version>
+    <version>0.9.1</version>
```
jnelsonjava, line 8: Not this version, the one on line 62. Should be 0.25.8 to match the SDK update.

imranmaj: Ah I see, the dependency

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
```diff
@@ -169,4 +172,39 @@ public void crudParticipantFile() throws Exception {
...
+        try {
+            userApi.getParticipantFile("rate-limit-test").execute();
+            fail("Download should have failed with a rate limit error");
+        } catch (BridgeSDKException e) {
+            assertEquals(429, e.getStatusCode());
+        }
+
+        // wait 5 seconds to let rate limit refresh so other tests and cleanup method
+        // can run
+        Thread.sleep(5000);
```
DwayneJengSage, line 208: Don't need to sleep, since Before is run for each test method, so each test will have its own user.

This is important because the tests already take a very long time, so we don't want to add unnecessary wait time.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
```diff
@@ -169,4 +172,39 @@ public void crudParticipantFile() throws Exception {
...
+        file = new ParticipantFile();
+        file.setMimeType("text/plain");
+        ParticipantFile upload = userApi.createParticipantFile("rate-limit-test", file).execute().body();
+        URL uploadUrl = new URL(upload.getUploadUrl());
+        HttpURLConnection connection = (HttpURLConnection) uploadUrl.openConnection();
+        connection.setDoOutput(true);
+        connection.setRequestMethod("PUT");
+        connection.setRequestProperty("Content-Type", "text/plain");
+        OutputStreamWriter out = new OutputStreamWriter(connection.getOutputStream(), StandardCharsets.US_ASCII);
+        out.write(StringUtils.repeat("A", 100)); // 100 B in ASCII
```
DwayneJengSage, line 187: Consider instead writing a 1kb file and reading it twice instead of writing 100b file and reading it 11 times. This means there's less overhead setting up the download.

DwayneJengSage (changes requested):
Overall looks good. A few minor optimizations.

src/test/java/org/sagebionetworks/bridge/sdk/integration/ParticipantFileTest.java
imranmaj: The issue I was having is with After, which calls getParticipantFiles in order to clean up the uploaded files immediately a rate limiting violation happens in the test, so that call is also rejected due to rate limiting

DwayneJengSage (approved):

PR #688, opened 2022-07-14
Change Exporter3Test to use bridge-base Config, not JavaSDK Config

DwayneJengSage:
JavaSDK Config cannot accept new config values through Env Vars without maintaining a hardcoded list of Env Vars in JavaSDK. This is especially important for maintaining Integ Tests.

This change switches to use bridge-base Config, which is consistent with Server, Worker, and Worker Integ tests. Since the vast majority of config in JavaSDK is actually used by Integ Tests, the goal is to eventually migrate all our configs to use bridge-base.


DwayneJengSage:
Dev is currently failing, so I'm going ahead and merging this to unblock Dev. Will follow up with Jack to make sure this is reviewed.

jnelsonjava (commented):
Looks good!

PR #694, opened 2022-09-02
Api to revert study to design phase (BRIDGE-3339)

jnelsonjava:
Testing ability for superadmins to move a study back to design outside of the typical study lifecycle.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyLifecycleTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyLifecycleTest.java
@@ -1,11 +1,9 @@
 package org.sagebionetworks.bridge.sdk.integration;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 import static org.sagebionetworks.bridge.rest.model.Role.STUDY_COORDINATOR;
 import static org.sagebionetworks.bridge.rest.model.Role.STUDY_DESIGNER;
-import static org.sagebionetworks.bridge.rest.model.Role.SUPERADMIN;
 import static org.sagebionetworks.bridge.rest.model.StudyPhase.ANALYSIS;
 import static org.sagebionetworks.bridge.rest.model.StudyPhase.COMPLETED;
 import static org.sagebionetworks.bridge.rest.model.StudyPhase.DESIGN;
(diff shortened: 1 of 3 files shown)


src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyLifecycleTest.java
```diff
@@ -41,13 +44,15 @@ public interface Supplier {
...
     private TestUser studyCoordinator;
+    private TestUser superadmin;
     private List<String> studiesToDelete;
     
     @Before
     public void before() throws IOException {
         studiesToDelete = new ArrayList<>();
         studyDesigner = TestUserHelper.createAndSignInUser(StudyLifecycleTest.class, false, STUDY_DESIGNER);
         studyCoordinator = TestUserHelper.createAndSignInUser(StudyLifecycleTest.class, false, STUDY_COORDINATOR);
+        superadmin = TestUserHelper.createAndSignInUser(StudyLifecycleTest.class, false, SUPERADMIN);
```
DwayneJengSage, line 55: We can get a superadmin by calling TestUserHelper.getSignedInAdmin().

That said, this won't work in prod, nor will the code that you wrote. The Integ Test account in Prod only has Admin privileges, not Superadmin, so it can neither call Superadmin APIs nor create Superadmins. This is by design, because we don't want to leak Superadmin credentials in Prod.

This is relevant because this integ test class is marked with @Category(IntegrationSmokeTest.class), which means it runs in Prod, and it will fail.

We will need to move this test to a separate class without the IntegrationSmokeTest annotation, so that it only runs in Dev and Staging.

src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyLifecycleTest.java
```diff
@@ -62,6 +67,9 @@ public void after() throws IOException {
         if (studyCoordinator != null) {
             studyCoordinator.signOutAndDeleteUser();
         }
+        if (superadmin != null) {
+            superadmin.signOutAndDeleteUser();
+        }
```
DwayneJengSage, line 72: When we switch to TestUserAdmin.getSignedInAdmin(), be sure to remove this. Otherwise, bad things happen.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyLifecycleTest.java
jnelsonjava: Removed!

src/test/java/org/sagebionetworks/bridge/sdk/integration/StudyLifecycleTest.java
jnelsonjava: That's a great catch. I just moved it to StudyTest which does not have the smoke test annotation.

DwayneJengSage (approved):

PR #695, opened 2022-09-20
Update integration tests to use mobile client assessment format

imranmaj:
https://sagebionetworks.jira.com/browse/BRIDGE-2964


DwayneJengSage (commented):
Looks good. Will need to bump the version on the JavaSDK and also bump the version here as well.

DwayneJengSage (approved):

PR #704, opened 2023-01-21
Clean out test accounts in InitListener

DwayneJengSage:


src/test/java/org/sagebionetworks/bridge/sdk/integration/InitListener.java
```diff
@@ -175,6 +178,30 @@ public void testRunStarted(Description description) throws Exception {
...
+        List<AccountSummary> accountSummaryList;
+        int numPages = 0;
+        do {
+            if (numPages >= 5) {
+                // Short-circuit to prevent an infinite loop.
+                LOG.error("Too many test accounts. Breaking out of account cleanup loop.");
+                break;
+            }
+
+            accountSummaryList = participantsApi.searchAccountSummaries(search).execute().body().getItems();
```
DwayneJengSage, line 194: Verify this is scoped to the current app.

DwayneJengSage: I have confirmed that this is scoped to the current app, so there is no risk of this deleting accounts outside of the test app.

jnelsonjava (approved):
Looks good

PR #707, opened 2023-02-08
Fix new enrollment alerts tests

imranmaj:
https://sagebionetworks.jira.com/browse/MTB-651

Re-enables new enrollment alert tests and adds an additional test for enrollments being added before an account is created. See [corresponding Bridge Server PR](https://github.com/Sage-Bionetworks/BridgeServer2/pull/645) for additional explanation.

All tests that previously failed now pass on my machine.


imranmaj:
Add test for AccountService.updateAccount creating alerts

DwayneJengSage (approved):

PR #723, opened 2023-10-04
DHP-1085 Add phase enum field to Assessment

nategbrown9:


DwayneJengSage (changes requested):
Additional test cases we should test (in AssessmentTest or AssessmentConfigTest):
* Set an assessment phase to REVIEW, attempting to update the AssessmentConfig will throw a BadRequestException.
* Set the study phase to RECRUITMENT, Assessments in that study should automatically be set to PUBLISHED. (Note, you'll need to revert the study phase back to DESIGN with a special API. See StudyTest.superadminCanRevertStudyPhaseToDesign())

DwayneJengSage (approved):

